PR #8, opened 2021-09-22
Only render guilds by urlName

### LINKED ISSUE #7: Filter for guilds when rendering by urlName
Right now [https://alpha.guild.xyz/agora](https://alpha.guild.xyz/agora) is a valid URL, however, agora is not a guild.
The `getStaticPaths` of `[guilds].tsx` only uses guilds, but because of ISR, on the `/agora` request, we still fetch the data.

cs-balazs:
Closes #7


cs-balazs:
Right now we can't request for only guilds by urlName, talked to @dezsim about this, the urlName endpoint will return an `isGuild` key, so we can avoid fetching all the guilds.

dovalid (approved):

dovalid (approved):

PR #132, opened 2022-01-06
Using LazyMotion

### LINKED ISSUE #122: Reduce bundle size: lazy load Framer Motion
use LazyMotion

BrickheadJohnny:
Closes #122


dovalid (approved):
We couldn't really cut any significant amount of the bundle size because Chakra already bundles these, but gonna merge it now so we get the benefits if they'll switch to LazyMotion too

PR #152, opened 2022-01-17
Fix machine warnings

### LINKED ISSUE #146: Fix machine warning
Fix console warning
`Machine given to "useMachine" has changed between renders. This is not supported and might lead to unexpected results.
Please make sure that you pass the same Machine as argument each time.`

cs-balazs:
Closes #146

### DIFF
--- a/src/hooks/useSubmit/useSubmit.ts
+++ b/src/hooks/useSubmit/useSubmit.ts
@@ -12,6 +12,8 @@ const useSubmit = <DataType, ResponseType>(
   fetch: (data: DataType) => Promise<ResponseType>,
   { onSuccess, onError }: Options<ResponseType> = {}
 ) => {
+  // xState does not support passing different objects on different renders,
+  // using a ref here, so we have the same object on all renders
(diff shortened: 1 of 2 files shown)


cs-balazs:
For some reason, this warning does not show up for me, but it did at drop.app, so I just did the same thing here.
Let me know if it fixed the warnings for you.

src/hooks/useSubmit/useSubmit.ts
```diff
@@ -11,7 +12,8 @@ const useSubmit = <DataType, ResponseType>(
   fetch: (data: DataType) => Promise<ResponseType>,
   { onSuccess, onError }: Options<ResponseType> = {}
 ) => {
-  const [state, send] = useMachine(createFetchMachine<DataType, ResponseType>(), {
+  const machine = useRef(createFetchMachine<DataType, ResponseType>())
```
dovalid, line 15: Could we add a comment here noting why it's necessary?

dovalid (approved):

PR #526, opened 2022-10-24
Update Unlock API

tchataigner:
This PR is meant to fix the behavior on the Unlock integration.

Previous Unlock subgraph integration was deprecated as per the documentation: https://docs.unlock-protocol.com/tools/subgraph/

Here we:
- Add some relevant networks
- Update the subgraphs endpoints


dovalid (approved):
Thanks tchataigner, awesome!

PR #552, opened 2022-11-04
Sound-xyz-integration

kef1rke:


src/requirements/Sound/SoundRequirement.tsx
```diff
@@ -0,0 +1,157 @@
+import { Img } from "@chakra-ui/react"
+import Link from "components/common/Link"
+import { RequirementComponentProps } from "requirements"
+import Requirement from "requirements/common/Requirement"
+import slugify from "slugify"
+import useSWRImmutable from "swr/immutable"
+
+const SoundRequirement = ({ requirement, ...rest }: RequirementComponentProps) => {
+  const { data: artistsData, isValidating: artistsLoading } = useSWRImmutable(
```
dovalid, line 9: shouldn't fetch if there's no requirement.data.id

src/requirements/Sound/SoundRequirement.tsx
```diff
@@ -0,0 +1,157 @@
...
+    `/api/sound-artists?searchQuery=${requirement.data?.id}`
+  )
+
+  const { data: songsData, isValidating: songsLoading } = useSWRImmutable(
+    `/api/sound-songs?id=${artistsData?.map((artist) => artist[0].id)}`
```
dovalid, line 14: shouldn't use map here, we need a value here not an array

src/requirements/Sound/SoundRequirement.tsx
```diff
@@ -0,0 +1,157 @@
...
+  )
+
+  const songImageUrl = songsData
+    ?.filter((title) => title[0].title == requirement.data.title)
```
dovalid, line 18: what's title[0].title? It's bad naming. But the endpoint shouldn't return an unnecessary multidimensional array in the first place

src/requirements/Sound/SoundRequirement.tsx
```diff
@@ -0,0 +1,157 @@
...
+            return songsData?.map((song) => song[0].image) ? (
+              <Img src={songImageUrl} />
+            ) : (
+              <Img src="/requirementLogos/sound.png" />
+            )
+          case "SOUND_TOP_COLLECTOR":
+            return artistsData?.map((artist) => artist[0].image) ? (
+              <Img src={artistsData?.map((artist) => artist[0].image)} />
+            ) : (
+              <Img src="/requirementLogos/sound.png" />
```
dovalid, line 45: should just have a `default: return ` at the end and conditional returns in earlier cases

src/requirements/Sound/SoundRequirement.tsx
```diff
@@ -0,0 +1,157 @@
...
+    .map((song) => song[0].image)
+
+  return (
+    <Requirement
+      image={(() => {
+        switch (requirement.type) {
+          case "SOUND_ARTIST":
+            return <Img src="/requirementLogos/sound.png" />
+
+          case "SOUND_ARTIST_BACKED":
```
dovalid, line 28: should just add a `case "SOUND_TOP_COLLECTOR":` line here since the logic is the same for them

src/requirements/Sound/SoundRequirement.tsx
```diff
@@ -0,0 +1,157 @@
...
+                <Link
+                  href={`https://www.sound.xyz/${requirement.data.id}`}
+                  isExternal
+                  fontWeight="medium"
+                  colorScheme={"blue"}
+                >
+                  {requirement.data.id}
+                </Link>
+                {` on `}
+                <Link
```
dovalid, line 141: since the `on Sound.xyz` link is always there at the end, could just move it after the switch statement instead of repeating it

src/requirements/Sound/components/SoundSupportArtist.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const {
+    control,
+    formState: { errors },
+  } = useFormContext()
+
+  const [search, setSearch] = useState(
+    useWatch({ name: `${baseFieldPath}.data.id` })
+  )
+
+  const { data: artistsData, isValidating: artistsLoading } = useSWRImmutable(
```
dovalid, line 22: the artist input and corresponding data fetching is repeated in every type's form, it should be abstracted to a SoundArtistSelect component

src/requirements/Sound/components/SoundSupportArtist.tsx
```diff
@@ -0,0 +1,81 @@
...
+import parseFromObject from "utils/parseFromObject"
+
+const SupportArtist = ({ baseFieldPath }: RequirementFormProps) => {
+  const {
+    control,
+    formState: { errors },
+  } = useFormContext()
+
+  const [search, setSearch] = useState(
+    useWatch({ name: `${baseFieldPath}.data.id` })
```
dovalid, line 19: it's weird to put an useWatch hook as useState's default value. Should get the field prop and put field.data.id here, that's why we have the field prop (in RequirementFormProps)

src/requirements/Sound/components/SoundSupportArtist.tsx
```diff
@@ -0,0 +1,81 @@
...
+              value={artistOptions?.find((option) => option.value === value)}
+              onChange={(newSelectedOption: SelectOption) => {
+                onChange(newSelectedOption?.value)
+              }}
+              onInputChange={(text, _) => {
+                setSearch(splitInput(text))
+              }}
+              isLoading={artistsLoading}
+              onBlur={onBlur}
+              // so restCount stays visible
```
dovalid, line 63: there's no restCount here, this is not needed

src/pages/api/sound-artists.ts
```diff
@@ -0,0 +1,52 @@
...
+                url
+              }
+            }
+          }
+        }
+      }`,
+    },
+  }).then((res) =>
+    res?.data?.search?.artists?.filter(
+      (artist) =>
```
dovalid, line 29: why is this JS filtering needed?

src/pages/api/sound-artists.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const handler: NextApiHandler = async (req, res) => {
+  const searchQuery = req.query.searchQuery.toString()
+
+  const soundArtists = await fetchArtists(searchQuery).then((data) =>
+    data?.map((info) => [
+      {
+        name: info?.name,
+        soundHandle: info?.soundHandle,
+        image: info?.user?.avatar?.url ?? "/requirementLogos/sound.png",
```
dovalid, line 43: fallback to `/requirementLogos/sound.png` is UI scope and it's already handled there so we shouldn't have this here

dovalid (changes requested):

src/pages/api/sound-artists.ts
kef1rke: it's needed because the api searches for every match, but i need only the soundHandle and artist name match

dovalid (approved):

PR #566, opened 2022-11-08
fetch token data from BE if it's missing

dovalid:


BrickheadJohnny:
@dovalid do we need this?

dovalid:
No, I don't think so. I'm not sure what has or what hasn't changed since then, but haven't encountered missing token data issues in a while, so it's good to close

PR #658, opened 2023-01-19
ControlledSelect component

### LINKED ISSUE #656: ControlledSelect component
We use the `StyledSelect` component wrapped in a `Controller` in many places, so we could make a `ControlledSelect` component which includes all the boilerplate code, and use that instead.

BrickheadJohnny:
Closes #656


BrickheadJohnny:
Note: I haven't updated the select in `SnapshotForm.tsx`, because we'll change a lot of stuff in that file in #596 - we'll need to update the selects there if this PR gets merged before that.

PR #659, opened 2023-01-19
nns resolve

balintligeti:


adamlaska (approved):

src/hooks/useLensProtocolName.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useWeb3React } from "@web3-react/core"
+import useSWRImmutable from "swr/immutable"
+
+const apolloClient = new ApolloClient({
+  uri: "https://api.lens.dev/",
+  cache: new InMemoryCache(),
+})
+
+const fetchLensProtocolName = (_, query) =>
+  apolloClient
```
BrickheadJohnny, line 11: I think we don't really need the Apollo client here, we could make a simple `POST` requests with our `fetcher` instead. Here's an example:

https://github.com/agoraxyz/guild.xyz/blob/main/src/requirements/Snapshot/hooks/useSpace.ts#L5-L18

BrickheadJohnny (changes requested):

src/hooks/useLensProtocolName.ts
```diff
@@ -0,0 +1,36 @@
...
+    .query({
+      query: gql(query),
+    })
+    .then((res) => res.data.profiles.items[0].handle)
+
+const useLensProtocolName = () => {
+  const { account } = useWeb3React()
+  const query = `
```
BrickheadJohnny, line 19: Minor impro: you could construct this query inside `fetchLensProtocolName` and pass only the `account` to that function, since that's the only variable data in it.

src/hooks/useUnstoppableName.ts
```diff
@@ -0,0 +1,26 @@
+import Resolution from "@unstoppabledomains/resolution"
+import { useWeb3React } from "@web3-react/core"
+import useSWRImmutable from "swr/immutable"
+
+const fetchUnstoppableName = (_, account, resolution) =>
+  resolution
+    .reverse(account)
+    .then((res) => res)
+    .catch()
```
dovalid, line 9: Why is this catch() needed here?

PR #666, opened 2023-01-20
feat: Buy requirements

### LINKED ISSUE #662: Buy requirements

BrickheadJohnny:
Closes #662


src/components/[guild]/Requirements/components/GuildCheckout/hooks/usePurchaseAsset.ts
```diff
@@ -0,0 +1,192 @@
...
+    source,
+    path,
+    tokenAddressPath,
+  ])
+
+  const generatedGetAssetsParams = useMemo(
+    () =>
+      purchaseAssetData ? generateGetAssetsParams(purchaseAssetData) : undefined,
+    [purchaseAssetData]
+  )
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/GuildCheckout/hooks/usePurchaseAsset.ts:140]: Can't we move calculating these to happen later, in the core functions? So we don't need these useMemos. As we already call `generateGetAssetsParams` in purchaseAsset, but pre-calculate it for useEstimateGasFee here. Ideally we'd just pass a purchaseAssetData containing variables in their original form, and convert them to BigNumber / whatever we need in the core functions with an abstracted function like generateGetAssetsParams

src/components/[guild]/Requirements/components/GuildCheckout/hooks/useAllowSpendingToken.ts
```diff
@@ -0,0 +1,27 @@
+import { MaxUint256 } from "@ethersproject/constants"
+import useContract from "hooks/useContract"
+import useShowErrorToast from "hooks/useShowErrorToast"
+import useSubmit from "hooks/useSubmit"
+import ERC20_ABI from "static/abis/erc20Abi.json"
+import useAllowance from "./useAllowance"
+
+const useAllowSpendingTokens = (tokenAddress: string, contract: string) => {
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/GuildCheckout/hooks/useAllowSpendingToken.ts:8]: I think it'd be nicer for useAllowance to contain this logic too, there's no need for a separate hook

src/components/[guild]/Requirements/components/GuildCheckout/components/FeeAndTotal.tsx
```diff
@@ -0,0 +1,139 @@
...
+} from "utils/guildCheckout/constants"
+import usePrice from "../hooks/usePrice"
+import usePurchaseAsset from "../hooks/usePurchaseAsset"
+import { useGuildCheckoutContext } from "./GuildCheckoutContex"
+
+const FeeAndTotal = (): JSX.Element => {
+  const { pickedCurrency, requirement } = useGuildCheckoutContext()
+
+  const textColor = useColorModeValue("gray.800", "gray.200")
+  const textAccentColor = useColorModeValue("black", "white")
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/GuildCheckout/components/FeeAndTotal.tsx:27]: Isn't `colorScheme="gray"` on the Text components and no colorScheme where textAccentColor is used sufficient?

src/components/[guild]/Requirements/components/GuildCheckout/components/PaymentCurrencyPicker/components/TokenInfo.tsx
```diff
@@ -0,0 +1,125 @@
...
+    data: { symbol: tokenSymbol, decimals, logoURI: tokenLogoURI },
+    error,
+    isValidating: isTokenDataLoading,
+  } = useTokenData(Chains[chainId], address)
+  const logoURI = ADDRESS_REGEX.test(address)
+    ? tokenLogoURI
+    : RPC[Chains[chainId]].nativeCurrency.logoURI
+  const symbol = ADDRESS_REGEX.test(address)
+    ? tokenSymbol
+    : RPC[Chains[chainId]].nativeCurrency.symbol
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/GuildCheckout/components/PaymentCurrencyPicker/components/TokenInfo.tsx:57]: This is already abstracted into useTokenData. The logo fallback should be too

src/components/[guild]/Requirements/components/GuildCheckout/components/PaymentCurrencyPicker/components/TokenInfo.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const { balance: tokenBalance, isLoading: isTokenBalanceLoading } =
+    useTokenBalance(address, chainId)
+  const balance = formatUnits(
+    (address === RPC[Chains[chainId]]?.nativeCurrency?.symbol
+      ? coinBalance
+      : tokenBalance) ?? "0",
+    address === RPC[Chains[chainId]]?.nativeCurrency?.symbol
+      ? RPC[Chains[chainId]]?.nativeCurrency?.decimals
+      : decimals ?? 18
+  )
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/GuildCheckout/components/PaymentCurrencyPicker/components/TokenInfo.tsx:70]: It would be nice if we could merge useCoinBalance & useTokenBalance into one hook that works for both (sidequest as per boy scout rule)

src/components/[guild]/Requirements/components/GuildCheckout/components/buttons/CloseModalButton.tsx
```diff
@@ -0,0 +1,18 @@
+import Button from "components/common/Button"
+import { useGuildCheckoutContext } from "../GuildCheckoutContex"
+
+const CloseModalButton = (): JSX.Element => {
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/GuildCheckout/components/buttons/CloseModalButton.tsx:4]: This is unused, I guess it got replaced by InfoModalFooter

src/components/[guild]/Requirements/components/GuildCheckout/components/buttons/ChooseCurrencyButton.tsx
```diff
@@ -0,0 +1,12 @@
+import Button from "components/common/Button"
+import CardMotionWrapper from "components/common/CardMotionWrapper"
+
+const ChooseCurrencyButton = (): JSX.Element => (
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/GuildCheckout/components/buttons/ChooseCurrencyButton.tsx:4]: This was confusing for me at first, would like it better if it'd be inlined in GuildCheckout, as the only button that's just a disabled placeholder

dovalid (changes requested):

src/components/[guild]/Requirements/components/GuildCheckout/hooks/usePurchaseAsset.ts
Comment (BrickheadJohnny) [reply src/components/[guild]/Requirements/components/GuildCheckout/hooks/usePurchaseAsset.ts]: @dovalid I ended up moving this logic (L92-L134) into the `generateGetAssetsParams` util function (see here f58e24a8b6fe6731d0336fe50c713a9824f9faca) and calling only that function inside the `usePurchaseAsset` hook, since we actually need to pass the generated params to the `useEstimateGasFee` hook (it's a general hook, which can be used with any contract, so I didn't want to put any TokenBuyer-specific logic there).

Let me know if you still think that we should avoid calling the `generateGetAssetsParams` function here.

Comment (dovalid) [reply src/components/[guild]/Requirements/components/GuildCheckout/hooks/usePurchaseAsset.ts]: It's perfect this way thanks!

PR #672, opened 2023-01-26
Payment as requirement

### LINKED ISSUE #671: Payment as requirement

BrickheadJohnny:
Closes #671


BrickheadJohnny:
## Changes

**In existing components:**

- Since we're already using the token picker select in 3 places, I've generalized the `TokenPicker` component, so we can use that everywhere
- The `BlockExplorer` component can be used with props now (we don't store the token address for `PAYMENT` requirements, but we fetch it on client side, so that's why I changed it this way)
- Added a `withLink` prop to the `Reward` component, so we can use it in the new payment form too
- `AddRequirement`: hiding `IsNegatedPicker` and `ModalFooter` if requirement type is `PAYMENT` + passing the `addRequirement` function to the requirement form components, so we can trigger it from inside the form (I'm open to suggestions for a better solution here!)
- `RequirementEditableCard`: hiding the edit button for `PAYMENT` requirements (this is temporary, the contract will probably support editable vaults in the future)

**Guild checkout refactor:**

- Removed the `GuildCheckout` component. We have 2 new components instead: `PurchaseRequirement` and `BuyPass`
- Made a `PaymentMethodButtons` component, so we can reuse the static "Pay with crypto" and "Pay with card" buttons
- Extracted the error collapse to its own component. I ended up leaving it out from the `BuyPass` modal, but at least `PurchaseRequirement` got a bit cleaner this way
- We can now pass children to the `TOSCheckbox` component, this way we can easily replace the text and reuse the component for both the purchase and the payment modal
- The `InfoModal` component is now composable: we can set its title and easily replace its content in progress/success/error state
- I've also made the `InProgress`, `Success`, and `TxError` components composable, so we can reuse them in both Guild checkout flows
- Removed the `tx` prop from the `TransactionLink` component, since we're storing it in the context, so we can use it from there
- Generalized the `TokenInfo` component, so we can use it for both checkout flows. Moved the purchase-related logic into `CurrencyListItem` which we're using only for the purchase flow (we'll only have 1 currency in the `PAYMENT` requirement, so we don't actually need a picker there)
- Moved the `PriceFallback` component to its own file, so we can reuse it
- Created a `useSubmitTransaction` hook, which contains the boilerplate code which we need when we use `useSubmit` for contract interactions inside `GuildCheckoutContext`
- Finally, renamed a couple of Guild checkout related components, so we can easily see which one is used for the purchase requirement or the buy pass flow

**Notes & TODO for later:**

- We could move the `useCoinById` and `useFeeInUSD` hooks to a more logical directory, I just haven't moved them because we might need to modify/move them in the "POAP as a reward" PR.
- There are some duplicate hooks now (e.g. `useWithdraw`) - those should be removed once we rework the POAP-related features.

dovalid (approved):

PR #673, opened 2023-01-27
Hotfix member count at discord role creation

ejayxyz:
Refactored the `useServerData` hook for the guildify component.

Removed `membersWithoutRole` property, because as far as I checked, we don't use.

### DIFF
--- a/src/hooks/useServerData.ts
+++ b/src/hooks/useServerData.ts
@@ -27,7 +27,6 @@ export type Role = {
 
 type ServerData = {
   serverIcon: string
-  membersWithoutRole: number
   serverName: string
   serverId: string
   categories: Category[]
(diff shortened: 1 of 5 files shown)


BrickheadJohnny (approved):

src/hooks/useServerData.ts
```diff
@@ -57,7 +57,7 @@ const useServerData = (
     shouldFetch
       ? [
           `/discord/server/${serverId}/${option?.memberCountDetails}`,
-          { method: "POST" },
+          { method: "GET" },
```
BrickheadJohnny, line 60: I think you can completely remove this option object here, since the default method is `GET` anyways. Then, you can also remove the square brackets around the fetch URL and just pass it as a single template string, so the code will be much cleaner.

ejayxyz: Ah, you probably right, but I just already merged it into main. We will fix it later, I guess.

PR #694, opened 2023-02-17
E2E tests

BrickheadJohnny:


cypress/e2e/0-platformless/1-cleanup.spec.ts
```diff
@@ -1,14 +1,15 @@
...
-    cy.visit(`/${Cypress.env("guildUrlName")}`, { failOnStatusCode: false })
+    cy.visit(
+      `/${Cypress.env("platformlessGuildUrlName")}-${Cypress.env("DEPLOYMENT_ID")}`,
+      {
+        failOnStatusCode: false,
+      }
+    )
   })
 
   it("cleans up test guild", () => {
```
cs-balazs, line 15: I think I had something similar when I was doing tests, but can't remember exactly why it is needed to check for 404/client-side error here, is this still required? If so, in case of 404 does the guild get cleaned up?
I think there used to be some backend inconsistency (or maybe something related to FE static generation(?))

cs-balazs (approved):

cypress/e2e/0-platformless/1-cleanup.spec.ts
BrickheadJohnny: yes, I believe you added it because sometimes the frontend will still show a 404 page, and then revalidates the guild on client-side 👍

cypress/support/commands.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import "@testing-library/cypress"
+import "cypress-file-upload"
```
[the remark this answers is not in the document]
BrickheadJohnny: Good point, thanks, will remove these deps!

PR #700, opened 2023-02-21
Guild Checkout improvements

BrickheadJohnny:


src/components/[guild]/Requirements/components/GuildCheckout/components/PurchaseFeeAndTotal.tsx
```diff
@@ -69,22 +69,28 @@ const PurchaseFeeAndTotal = (): JSX.Element => {
...
-                  : (calculatedGasFee ?? 0 + guildFeeInSellToken ?? 0)?.toFixed(
-                      3
-                    )}{" "}
-                {symbol}
-              </>
-            ) : (
-              "Choose currency"
+          <Skeleton
+            isLoaded={Boolean(
+              pickedCurrency && calculatedGasFee && guildFeeInSellToken
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/GuildCheckout/components/PurchaseFeeAndTotal.tsx:74]: Is there no state when it appears loading because one of the conditions is falsy, but it's not actually loading just ran into an error?

Comment (BrickheadJohnny) [reply src/components/[guild]/Requirements/components/GuildCheckout/components/PurchaseFeeAndTotal.tsx]: Good point, fixed!

PR #707, opened 2023-02-24
Visibility

cs-balazs:
# Visibility

  - Roles, requirements, and rewards can have a visibility value:
    - `PUBLIC`: Everyone can see it
    - `PRIVATE`: Only holders of the role can see it
    - `HIDDEN`: Only admins of the guild can see it
  - These values need to be modifiable in the role creation, and role edit drawers
  - If we hide something from a user (for example because of not being an admin), we need to display the fact that *some* role/requirement/reward is hidden
  - With this feature it is possible that a user can't see any requirements, this also ha to be handled
  - BE now uses optional authentication on many endpoints, so we need to add the logic to handle that

## Implementation details
  - 🪝 `useSWRWithOptionalAuth` for handling optional auth on an SWR call level. Works similarly to the previous `useGuild`. Makes an immutable call as an unauthenticated request (this serves both as a fallback, and the actual data, when the user is not authenticated), and if the user is authenticated, makes a configurable request with authentication
    - Also exports a `mutateOptionalAuthSWRKey` so we handle this kind of mutation in one place
  - ⚛️ `SetVisibility` for setting visibility value on a given RRR entity
  - ⚛️ `Visibility` for displaying these attributes to the admin
  - `RadioOption` has been extended
    - `disabled` can now be a `boolean` for the case when we don't need to display a `Tag`
    - `tooltipLabel`: If exists, the option is wrapped in a `Tooltip` with this label
    - `RightComponent`: If exists, gets displayed on the right side of the `RadioOption`
  - `fetcher` has been extended to handle optional authentication

## Additional information
  - Related BE PR: agoraxyz/guild-backend#708
  - This branch currently points to the dev api, this needs to be changed to prod before merging


src/components/[guild]/JoinModal/hooks/useJoin.tsx
```diff
@@ -80,9 +80,11 @@ const useJoin = (onSuccess?: () => void) => {
 
       addDatadogAction(`Successfully joined a guild`)
 
-      mutate(`/user/membership/${account}`)
-      // show user in guild's members
-      mutate(`/guild/${router.query.guild}`)
+      setTimeout(() => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/JoinModal/hooks/useJoin.tsx:83]: Why do we need this setTimeout here? Do we need to wait for the 2 other mutations above (in that case we could maybe await those)?

src/components/[guild]/Requirements/RoleRequirements.tsx
```diff
@@ -14,11 +14,18 @@ const VIRTUAL_LIST_REQUIREMENT_LIMIT = 10
 const PARENT_PADDING = "var(--chakra-space-5)"
 
 const RoleRequirements = ({ role }: Props) => {
-  const isVirtualList = role.requirements?.length > VIRTUAL_LIST_REQUIREMENT_LIMIT
-  const sliceIndex = (role.requirements?.length ?? 0) - 3
-  const shownRequirements = (role.requirements ?? []).slice(0, 3)
+  const requirements = role.hiddenRequirements
+    ? [
+        ...role.requirements,
+        { isHidden: true, type: "FREE" } as unknown as Requirement,
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/RoleRequirements.tsx:20]: We could maybe add the `isHidden?: boolean` prop to the `Requirement` type, and then we could just cast this object to `Requirement` type directly?

src/components/[guild]/Requirements/RoleRequirements.tsx
```diff
@@ -14,11 +14,18 @@ const VIRTUAL_LIST_REQUIREMENT_LIMIT = 10
...
+      ]
+    : role.requirements
+
+  const isVirtualList = requirements?.length > VIRTUAL_LIST_REQUIREMENT_LIMIT
+  const sliceIndex = (requirements?.length ?? 0) - 3
+  const shownRequirements = (requirements ?? []).slice(0, 3)
   const hiddenRequirements =
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/RoleRequirements.tsx:27]: TODO for myself: I think I'll rename this variable, because the `hiddenRequirements` name might be confusing now that we have requirements which should be actually hidden (and not just hidden until the user doesn't click on the "and x more" button 😄

src/components/[guild]/Requirements/components/RequirementDisplayComponent.tsx
```diff
@@ -17,6 +17,14 @@ const RequirementDisplayComponent = ({
   rightElement,
   ...rest
 }: Props) => {
+  if ((requirement as any).isHidden) {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/components/RequirementDisplayComponent.tsx:20]: If we add the `isHidden` prop to `RequirementType`, we can remove the typecasting here too. Or is there a reason why you decided not to add it to types? 👀

src/components/[guild]/JoinModal/hooks/useJoin.tsx
Comment (cs-balazs) [reply src/components/[guild]/JoinModal/hooks/useJoin.tsx:83]: I think it was needed because of read replica lag. The mutations are needed because we might see some additional RRR-s after we joined a guild, but when I tried without timeout, I received the same older data

src/components/[guild]/Requirements/components/RequirementDisplayComponent.tsx
Comment (cs-balazs) [reply src/components/[guild]/Requirements/components/RequirementDisplayComponent.tsx:20]: No specific reason, I guess I just didn't want to add it because it's only used here, and not directly related to a requirement entity. But I tried it locally, and I think you are right, we should add it to the type instead, with a comment describing, that this is not really a part of a requirement

PR #715, opened 2023-03-04
Connect social accounts

balintligeti:


src/components/[guild]/JoinModal/hooks/useConnectPlatform.ts
```diff
@@ -63,6 +61,10 @@ const useConnectPlatform = (
     onSuccess: () => {
       addDatadogAction("Successfully connected 3rd party account")
       mutateUser()
+      toast({
```
Comment (dovalid) [inline src/components/[guild]/JoinModal/hooks/useConnectPlatform.ts:64]: this shouldn't be abstracted here, but passed as onSuccess in LinkMoreSocialAccount, like in other components that uses the hook (eg ConnectRequirementPlatformButton). So we can have different toasts for different cases (now it shows double toasts in other cases)

src/components/common/Layout/components/Account/components/AccountModal/components/LinkMoreSocialAccount.tsx
```diff
@@ -0,0 +1,56 @@
...
+import platforms from "platforms/platforms"
+import { useEffect } from "react"
+import { PlatformName } from "types"
+import capitalize from "utils/capitalize"
+
+type Props = {
+  platformName: string
+}
+
+const LinkMoreSocialAccount = ({ platformName }: Props): JSX.Element => {
```
dovalid, line 20: LinkMoreSocialAccount sounds strange to me, I'd suggest LinkNewSocialAccount

src/components/common/Layout/components/Account/components/AccountModal/components/LinkMoreSocialAccount.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { isOpen, onOpen, onClose } = useDisclosure()
+
+  const { onConnect, isLoading, response } = useConnectPlatform(
+    platformName as PlatformName,
+    onClose
+  )
+  useEffect(() => onOpen(), [platformName])
```
dovalid, line 27: we should add {defaultIsOpen: true} to useDisclosure instead of this. If you've tried that and it didn't work, we should add a comment for a hacky useEffect like this

src/components/common/Layout/components/Account/components/AccountModal/components/LinkMoreSocialAccount.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    <Collapse in={isOpen} animateOpacity>
```
dovalid, line 30: minor note, animateOpacity is true by default, we should only specify it if we wan't to override it to be false

src/components/common/Layout/components/Account/components/AccountModal/components/LinkMoreSocialAccount.tsx
```diff
@@ -0,0 +1,56 @@
...
+          icon={
+            <Icon as={platforms[platformName]?.icon} boxSize={4} color="white" />
+          }
+          boxSize={8}
+          bgColor={`${platforms[platformName]?.colorScheme}.500`}
+        />
+        <Text fontWeight="semibold">{capitalize(platformName.toLowerCase())}</Text>
+
+        <Button
+          isLoading={isLoading}
```
dovalid, line 42: note: isLoading was custom handled before at rightIcon, isDisabled and JSX children, there's a prop for that that handles these automatically, I've changed this in my UI commit

dovalid (changes requested):

src/components/common/Layout/components/Account/components/AccountModal/components/LinkMoreSocialAccount.tsx
balintligeti: {defaultIsOpen: true} doesn't works, when i disconnect an account, gonna add a comment

PR #717, opened 2023-03-06
UI snap tasks

balintligeti:


src/pages/explorer.tsx
```diff
@@ -112,8 +116,49 @@ const Page = ({ guilds: guildsInitial }: Props): JSX.Element => {
     setColorMode("dark")
   }, [])
 
+  const router = useRouter()
+  const [isRouteChanges, setisRouteChanges] = useState(false)
+  const { colorMode } = useColorMode()
+
+  useEffect(() => {
+    const handleRouteChangeStart = () => {
+      setisRouteChanges(true)
```
dovalid, line 125: we should set the state to true only after a let's say 500ms setTimeout, otherways the navigation feels instant and it's better not to show any loading animation

dovalid (changes requested):

PR #729, opened 2023-03-14
Activity log

BrickheadJohnny:
### Notes for the reviewer
- I decided to make an `ActivityLogFiltersContext`, where we store the active filters with their values, and this context also manages the URL search params (e.g. calls `router.push` if needed). This way we can easily update filters both in the `FiltersInput` and `DateRangeInput` components
- I also made an `ActivityLogActionContext`, so we don't need to pass props to every component inside `AuditLogAction`, we just provide all the necessary data in this context
- (The `ActionIcon`, `ActionLabel`, and `ActivityLogChildAction` components became pretty clear thanks to this)
- I also made `ClickableUserTag`, `ClickableRoleTag`, `ClickableRewardTag` components, so the user can easily filter the logs by clicking on these tags, but as we discussed it earlier, I'll modify these once the CRM is live, so we can "connect" the Activity Log and the CRM with these tags
- I also left a couple of comments in this PR, please check those ;)

### Questions
- I wasn't sure how to solve the issue with long lines in the `MoreActions` component. I implemented a horizontal scroll feature, which would work "okay" on smaller screens, but not sure if it is a good UX or not. Alternatively we could do something like in the `RoleRequirements` component, where we calculate row heights in the virtual list on the fly (we could enable line wrapping in this case)


src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag.tsx
```diff
@@ -0,0 +1,397 @@
...
+        {...rootProps}
+      >
+        <HStack {...controlProps}>
+          <Text as="span" fontSize="sm" fontWeight="bold">
+            {filterNames[filter]}:
+          </Text>
+
+          {value ? (
+            <>
+              {(() => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag.tsx:144]: We could extract this logic to its own component, but I felt like it's not that bad this way. Let me know what's your opinion on this!

src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag.tsx
```diff
@@ -0,0 +1,397 @@
...
+
+      {["roleId", "rolePlatformId", "action"].includes(filter) && (
+        <Dropdown
+          {...{
+            ...positionerProps,
+            style: { ...positionerProps.style, ...positionerStyle },
+          }}
+        >
+          <Stack spacing={0} {...contentProps}>
+            {(() => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag.tsx:285]: Same here

src/components/[guild]/activity/ActivityLogFiltersBar/components/FiltersInput.tsx
```diff
@@ -0,0 +1,253 @@
...
+import {
+  isSupportedQueryParam,
+  SupportedQueryParam,
+  useActivityLogFilters,
+} from "./ActivityLogFiltersContext"
+import Dropdown from "./Dropdown"
+import FilterTag from "./FilterTag"
+import Suggestion from "./Suggestion"
+
+const FILTERS_TO_RENDER: SupportedQueryParam[] = [
```
Comment (dovalid) [inline src/components/[guild]/activity/ActivityLogFiltersBar/components/FiltersInput.tsx:25]: Why is this needed instead of just using SUPPORTED_SEARCH_OPTIONS?

src/components/[guild]/activity/ActivityLogFiltersBar/components/FiltersInput.tsx
```diff
@@ -0,0 +1,253 @@
...
+  "action",
+  "userId",
+  "roleId",
+  "rolePlatformId",
+]
+
+const getPositionerCSSVariables = (
```
Comment (dovalid) [inline src/components/[guild]/activity/ActivityLogFiltersBar/components/FiltersInput.tsx:32]: Maybe this whole function is unnecessary and we could just use getComputedStyle?

src/components/[guild]/activity/ActivityLogAction/components/MoreActions.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const { isOpen, onOpen, onClose } = useDisclosure()
+  const numberOfAdditionalActions = actions.length - displayedActionCount
+
+  const Row = memo(({ index, style }: { index: number; style: CSSProperties }) => {
+    const action = actions[index]
+
+    return (
+      <ListItem style={style}>
+        <ActivityLogActionProvider key={action.id} action={action}>
+          <Box
```
Comment (dovalid) [inline src/components/[guild]/activity/ActivityLogAction/components/MoreActions.tsx:41]: I think we should make the whole container horizontally scrollable, not just the lines by themeselves, and that way this solution is sufficient

src/components/[guild]/activity/ActivityLogFiltersBar/components/FiltersInput.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/activity/ActivityLogFiltersBar/components/FiltersInput.tsx:32]: Yup, I tried that already, but we don't actually need the whole style object (it doesn't really work if we pass it down to the `FilterTag` component), and we only need the `transform` prop when its value is not `"none"`, so I decided to write a short function which retrieves only the necessary props from computed styles. But let me know if you can think about a cleaner solution, because I'm not 100% satisfied with this one either :D

src/components/[guild]/activity/ActivityLogFiltersBar/components/FiltersInput.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/activity/ActivityLogFiltersBar/components/FiltersInput.tsx]: Good point, fixed in 1f2c67a5892e0ed31913f387a757fbab3c240e0b

src/components/[guild]/activity/ActivityLogAction/components/MoreActions.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/activity/ActivityLogAction/components/MoreActions.tsx]: Actually, after experimenting with multiple solution, I felt like making a variable size list is a better solution, so I implemented it in 1c2a33bf062810f2b8e2122bc040d3fcb7c0db7f (note: the list item heights don't update on window resize events, but I thought that's a rare edge case)

src/components/[guild]/activity/ActivityLogAction/components/RewardTag.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { useActivityLog } from "../../ActivityLogContext"
+import { useActivityLogFilters } from "../../ActivityLogFiltersBar/components/ActivityLogFiltersContext"
+
+type Props = {
+  name?: string
+  icon?: (props: IconProps) => JSX.Element
+  colorScheme?: ChakraProps["color"]
+} & Omit<TagProps, "colorScheme">
+
+const RewardTag = forwardRef<Props, "span">(
```
Comment (dovalid) [inline src/components/[guild]/activity/ActivityLogAction/components/RewardTag.tsx:22]: Why is the abstraction so general, instead of just needing the type and id, and handling the rest? Providing data for the current props from platforms.ts (and a `platformId === PlatformType.DISCORD` condition) seems to be repeated 3 times now instead of just being here, but I might be missing something

Comment (BrickheadJohnny) [reply src/components/[guild]/activity/ActivityLogAction/components/RewardTag.tsx:16]: Oh, nice catch, refactored all these tags in e61a83ecf54aa148b5afb59f9775b4766e259754

dovalid (approved):

PR #733, opened 2023-03-16
Popup messaging rework

cs-balazs:
# OAuth popup messaging rework

This PR aims to make the messaging with OAuth popups cleaner and less prone to errors, and to add support for cases where the auth window rewrites the main window (like in Trust, and in Coinbase wallets)

## Implementation
- Uses `BroadcastChannel` instead of `localStorage` for communication
- In the case when we don't receive a confirmation from the main window (Trust & Coinbase), we set the `authData` to `localStorage`, and redirect the user to the opener's URL. There we check the `localStorage`, and make connections if necessary
   - This is the reason, why `useConnectPlatform` has been modified
- Small `usePopupWindow` change to support passing URL in `onOpen` instead of the hook itself
  - With this, the CSRF token handling is much cleaner, since we don't need to construct the token in the top-level of the hook, we just generate it in the function
- `useOauthPopupWindow` now also takes a `PlatformName` param. This is needed for the case of Trust & Coinbase wallets, since we need to know, which platform the `authData` from the `localStorage` is for
- I've changed the encoding of the CSRF token from hex to base64. This way we represent the same 32 bytes in a shorter string, than in hex. This is important, because for example Twitter limits the maximum length of a redirect URL, and just strips the end if it is too long, which we interpret as an invalid CSRF token
- DataDog events:
  - Error: `OAuth - Unexpected error` (`{ error: unknown }`): If something unexpected happens in the `/oauth` page.
  - Error: `OAuth - Invalid CSRF token` (`{ received: string, expected: string }`): If the CSRF tokens don't match
  - Action: `OAuth - Message confirmation timed out`: If the popup didn't receive the confirmation from the main window in time (Trust & Coinbase)

The branch might look like it made things more complicated, but it is mostly because of the added support for Trust & Coinbase (and other browsers where the popup overwrites the main window). The state of the branch in this commit (https://github.com/agoraxyz/guild.xyz/pull/733/commits/61e71ba3e91c0c2527e92dc8afda5f025c5d00e4) shows how the communication itself got simpler


src/components/[guild]/JoinModal/hooks/useConnectPlatform.ts
```diff
@@ -28,14 +28,40 @@ const useConnectPlatform = (
...
   const { onOpen, authData, isAuthenticating, ...rest } =
     platformAuthHooks[platform]?.() ?? {}
   const prevAuthData = usePrevious(authData)
 
+  const { onSubmit, isLoading, response } = useConnect(onSuccess)
+
+  useEffect(() => {
+    // couldn't prevent spamming requests without all these three conditions
+    if (!platformUsers || !authData || prevAuthData) return
+    // const alreadyConnected = platformUsers.some(
```
Comment (BrickheadJohnny) [inline src/components/[guild]/JoinModal/hooks/useConnectPlatform.ts:41]: I think we could delete these lines, they're commented since months

src/components/[guild]/JoinModal/hooks/useOauthPopupWindow.ts
```diff
@@ -48,129 +18,116 @@ type OAuthOptions = {
...
-        mutateCSRFToken(randomBytes(16).toString("hex"))
-      })
-  }, [windowInstance])
+        resolve()
+      }
+    })
+
+    onOpen(
+      `${url}?${Object.entries(oauthOptions)
+        .map(([key, value]) => `${key}=${value}`)
```
Comment (BrickheadJohnny) [inline src/components/[guild]/JoinModal/hooks/useOauthPopupWindow.ts:103]: I think we could use URLSearchParams here to make it a bit cleaner, maybe something like this?

```ts
const searchParams = new URLSearchParams({
  ...oauthOptions,
  redirect_uri: redirectUri,
  state: csrfToken
}).toString()

onOpen(`${url}?${searchParams}`)
```

Comment (cs-balazs) [reply src/components/[guild]/JoinModal/hooks/useOauthPopupWindow.ts:103]: Thank you, this is indeed much cleaner!

src/components/[guild]/JoinModal/hooks/useConnectPlatform.ts
Comment (cs-balazs) [reply src/components/[guild]/JoinModal/hooks/useConnectPlatform.ts]: Removed it in https://github.com/agoraxyz/guild.xyz/pull/733/commits/405621046331f229e6a122e75b7e0d40373b4c72 👍

src/components/[guild]/JoinModal/hooks/useOauthPopupWindow.ts
Comment (cs-balazs) [reply src/components/[guild]/JoinModal/hooks/useOauthPopupWindow.ts]: Made this change in https://github.com/agoraxyz/guild.xyz/pull/733/commits/ddef7ec5acb3c516afce2a54bd13c8af08132be3

dovalid (approved):

PR #739, opened 2023-03-23
Collapse accessed roles

dovalid:
also includes:
- refactor & perf (RoleRequirements): extract VirtualRequirements logic into separate, memoized component
- UI (AccessIndicator): change "you have access" tag to "join to get access" button if not member yet
- refactor (AccessIndicator): conditional render to different places on mobile and desktop instead of changing placement with CSS
- refactor (PoapAccessIndicator): abstract expired & not yet active state
- UI (RoleCard): finetune centering of header elements
- UI (RoleCard): move Visibility below title when there's not enough horizontal space
- copy (MemberCount): show compact numbers


src/pages/_app.tsx
```diff
@@ -14,6 +14,11 @@ import { Fragment, useEffect, useState } from "react"
 import { SWRConfig } from "swr"
 import "theme/custom-scrollbar.css"
 import fetcher from "utils/fetcher"
+/**
+ * Polyfill HTML inert property for Firefox support:
+ * https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert#browser_compatibility
+ */
+require("wicg-inert")
```
BrickheadJohnny, line 21: Could we use a simple `import` here (just for consistency)

src/components/[guild]/RoleCard/RoleCard.tsx
```diff
@@ -54,92 +83,165 @@ const RoleCard = memo(({ role }: Props) => {
...
-                overflowWrap={"break-word"}
+      <Collapse in={isOpen} startingHeight={collapsedHeight}>
+        <SimpleGrid columns={{ base: 1, md: 2 }}>
+          <Flex
+            direction="column"
+            p={5}
+            borderRightWidth={{ base: 0, md: 1 }}
+            borderRightColor={
+              !isOpen
+                ? "transparent"
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RoleCard/RoleCard.tsx:94]: I think we could use the `useColorModeValue` hook in this file, so we could remove the double ternary here (it's a bit hard to parse, at least for me, haha)

So I'd do something like this:
```ts
const requirementsListBgColor = useColorModeValue("gray.50", "blackAlpha.300")
const roleDetailsBorderColor = useColorModeValue("gray.200", "gray.600")
  ```

BrickheadJohnny (approved):
Left 2 small comments, but otherwise looks good to me! Also tested it a couple of times, it looks & feels really smooth, love it!

src/pages/_app.tsx
dovalid: Oh right, I've only tried `import wicg-inert from "wicg-inert"` and that got removed by VS Code because of being unused, so went with a require. But simply `import "wicg-inert"` works, changed it, thanks!

src/components/[guild]/RoleCard/RoleCard.tsx
Comment (dovalid) [reply src/components/[guild]/RoleCard/RoleCard.tsx]: True, I agree, switched to invidual hooks

PR #746, opened 2023-03-27
New payment contract

dovalid:


BrickheadJohnny (changes requested):
I only found one possible issue in the code, so I requested a change. I also tested on the Vercel deployment, and it looks almost perfect to me, just three more things:

- When I set up a payment for a POAP, I can see the "edit" button next to it: if I click on it, I get a client-side error. I think we should just hide that button, since payment requirements aren't editable for now
- It would be great to calculate the actual withdrawable amount and display it on the withdraw button
- I think we should display an alert if the user tries to close the vault setup modal, to avoid accidental user errors
- If I mint the POAP in the access hub, the success modal disappears instantly, so I can't see my mint link

Not sure if these latter three are out of the scope of this PR or not, but it would be great to fix these two issues at some point.

PR #793, opened 2023-05-08
Update to SWR 2

dovalid:


BrickheadJohnny (approved):
Awesome!
I just remembered that we've implemented a custom logic in the `usePrice` hook so we can always return the previous data while we're revalidating, but now I removed that, because we can use the new `keepPreviousData` option for that.

PR #803, opened 2023-05-24
V2 - Paginated explorer

cs-balazs:
# V2 - Paginated explorer

This PR aims to reimplement guilds fetching in the explorer page with the v2 paginated endpoint. User's guilds can now be obtained from this endpoint using optional authentication, so this SWR call has been modified as well
The v2 endpoint currently doesn't return roles, only rolesCounts. This is to further reduce load on the DB (we can improve on this in the future if necessary). Platforms is also not returned for the same reason

### DIFF
--- a/src/components/explorer/YourGuilds.tsx
+++ b/src/components/explorer/YourGuilds.tsx
@@ -10,9 +10,9 @@ import GuildCardsGrid from "components/explorer/GuildCardsGrid"
 import useMemberships, {
   Memberships,
 } from "components/explorer/hooks/useMemberships"
+import useSWRWithOptionalAuth from "hooks/useSWRWithOptionalAuth"
 import { Plus, Wallet } from "phosphor-react"
 import { useEffect, useState } from "react"
-import useSWR from "swr"
 import { GuildBase } from "types"
 
 type Props = {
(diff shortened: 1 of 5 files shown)


src/components/explorer/YourGuilds.tsx
```diff
@@ -40,13 +40,8 @@ const YourGuilds = ({ guildsInitial }: Props) => {
...
 
-  /**
-   * Fetching all guilds without search params to filter user's guilds out of it.
-   * `?order=members` is the default for the other filtered request too, so only one
-   * request will be sent. There'll be a separate endpoint for it in the future
-   */
-  const { data: allGuilds, isLoading: isGuildsLoading } = useSWR(
-    `/guild?order=members`,
+  const { data: allGuilds, isLoading: isGuildsLoading } = useSWRWithOptionalAuth(
+    `/guild?`, // ? is included, so the request hits v2Replacer in fetcher
```
BrickheadJohnny, line 44: Not that important, but could you please change `IntercomProvider` too, so it uses this cache key?

src/utils/fetcher.ts
```diff
@@ -10,6 +10,24 @@ const SIG_HEADER_NAME = "x-guild-sig"
...
+const v2Replacer = (endpoint: string, options: Record<string, any>) => {
+  // if (options.method?.toLowerCase() === "post" && endpoint.includes("/v1/guild")) {
+  //   return endpoint.replace("/v1/guild", "/v2/guild/with-roles")
+  // }
+  // if (options.method?.toLowerCase() === "get" && endpoint.includes("/v1/guild/")) {
+  //   return endpoint.replace("/v1/guild/", "/v2/guild/guild-page/")
+  // }
+
+  if (
+    (options.method?.toLowerCase() ?? "get") === "get" &&
```
BrickheadJohnny, line 22: Do we actually need this line? 👀 I guess we won't post/put/delete to a `/guild?` endpoint anyways, and "get" === "get" will always be true? But I might miss something here haha

cs-balazs: Oh, right, since the `?` is there, checking the method doesn't seem necessary 👍 
The `?? "get"` is just there, because for GETs, we don't specify it explicitly, and it is just `undefined`. So `undefined` means `"get"` here

src/components/explorer/YourGuilds.tsx
cs-balazs: Thanks! Fixed in https://github.com/agoraxyz/guild.xyz/pull/803/commits/1af7a4ad54cc1257ae2172694a056482994be918 if that is what you meant 👍

src/utils/fetcher.ts
cs-balazs: Removed in https://github.com/agoraxyz/guild.xyz/pull/803/commits/eb301d4a52c0170c87384e4677c59880cca3b045 👍

BrickheadJohnny (approved):

PR #804, opened 2023-05-24
fix: color mode initial flash

dovalid:


BrickheadJohnny:
Chakra UI [issue](https://v2.chakra-ui.com/docs/styled-system/color-mode#color-mode-flash-issue), we couldn't really find a solution for this.

PR #811, opened 2023-05-30
UI: sync role description & requirements height

dovalid:
Example guild: traderjoe


BrickheadJohnny (approved):
Awesome! I added `descriptionRef` types and I modified the toggle logic a bit in my latest commit, because it was really strange to me that the description closed when I clicked on it (e.g. I tried to click on a link in it or to select a part of the text). With this solution, it'll still close if the user closes the requirement list, but it won't close if they click on the description. I'm not sure what's your opinion of this one, but if you don't like it this way, you only need to revert the latest commit ;)

PR #825, opened 2023-06-12
address merge: remove `localStorage` usage

cs-balazs:
# address merge: remove `localStorage` usage

The usage of `localStorage` causes some invalid states. I think it makes sense not to use `localStorage` for this action, and instead use an SWR state, or a simple state in a context. In this PR, I've tried the latter
Seems to work fine, the problem is, that I don't remember why we implemented it with `localStorage` originally

### DIFF
--- a/src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
+++ b/src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
@@ -40,32 +40,25 @@ type Props = {
   onOpen: () => void
 }
 
-const fetchShouldLinkToUser = async ([_, userId]) => {
-  try {
-    const { id: userIdToConnectTo } = JSON.parse(
-      window.localStorage.getItem("userId")
-    )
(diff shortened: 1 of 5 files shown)


src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
```diff
@@ -113,8 +107,16 @@ const WalletSelectorModal = ({ isOpen, onClose, onOpen }: Props): JSX.Element =>
   }, [keyPair, ready, router])
 
   const { data: shouldLinkToUser } = useSWR(
```
dovalid, line 109: sidequest: would be nice to extract it with `fetchShouldLinkToUser` into a `useShouldLinkToUser` hook so the logic is coupled there and not flooded here

cs-balazs: Sidequest completed :ayy:
https://github.com/agoraxyz/guild.xyz/pull/825/commits/22d533bef104baf0036f9ab7c018e15e12306689

PR #827, opened 2023-06-13
Account modal redesign

dovalid:


BrickheadJohnny (approved):
Redesigned the pins section too while I was here, but the rest looks fantastic in my opinion, approved!

PR #835, opened 2023-06-15
CRM

dovalid:
general abstraction changes:
- useToastWithButton now accepts a second button too (needed for the new shared socials toast's "Later" button)
- useScrollEffect now accepts an element prop, that defaults to document (so we can use it for the table's scroll container)

filter & sort functions are commented out because we've switched to server side filtering/sorting, maybe we could delete them altogether


src/components/[guild]/crm/IdentitiesSearch.tsx
```diff
@@ -0,0 +1,76 @@
...
+  InputGroup,
+  InputLeftElement,
+  InputRightElement,
+} from "@chakra-ui/react"
+import { Column } from "@tanstack/react-table"
+import useDebouncedState from "hooks/useDebouncedState"
+import { MagnifyingGlass } from "phosphor-react"
+import { useEffect, useRef, useState } from "react"
+
+// export const identitiesFilter: FilterFn<any> = (row, columnId, value, addMeta) => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/crm/IdentitiesSearch.tsx:14]: We can remove this too as you mentioned

src/components/[guild]/crm/IdentitiesSearch.tsx
```diff
@@ -0,0 +1,76 @@
...
+//   return false
+// }
+
+type Props = {
+  column: Column<any>
+}
+
+const IdentitiesSearch = ({ column }: Props) => {
+  const inputRef = useRef<HTMLInputElement>()
+  const [localValue, setLocalValue] = useState(column.getFilterValue() as string)
```
Comment (BrickheadJohnny) [inline src/components/[guild]/crm/IdentitiesSearch.tsx:35]: I tried to find a solution to type these properly, but seems like the `getFilterValue` function always returns `unknown`, right? In that case, I can't really find a better solution than this.

src/pages/[guild]/members.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const [sorting, setSorting] = useState(() => parseSortingFromQuery(router.query))
+  const [rowSelection, setRowSelection] = useState({})
+
+  const queryString = useMemo(
+    () => buildQueryStringFromState(columnFilters, sorting),
+    [columnFilters, sorting]
+  )
+
+  useEffect(() => {
+    window.history.pushState("", "", `?${queryString}`)
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/members.tsx:52]: Can't we use next router here? I think the browser navigation doesn't work properly this way. (at least if I set some filters on the members page, then visit another page, e.g. "Events", then click on the back button in the browser, the app won't navigate back to the "Members" page)

src/pages/[guild]/members.tsx
```diff
@@ -0,0 +1,227 @@
...
+            }}
+            colorScheme="primary"
+            mt="2px"
+          />
+        ),
+      },
+      columnHelper.accessor((row) => row, {
+        id: "identity",
+        size: 210,
+        // filterFn: identitiesFilter,
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/members.tsx:87]: We can remove these as you wrote in the PR description

src/components/[guild]/crm/useMembers.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+}
+
+const LIMIT = 50
+
+const useMembers = (queryString) => {
+  const { id } = useGuild()
+  const [hasReachedTheEnd, setHasReachedTheEnd] = useState(true)
+
+  const getKey = useCallback(
```
Comment (BrickheadJohnny) [inline src/components/[guild]/crm/useMembers.tsx:31]: We should return null if the user's keyPair isn't ready, so we don't get a signing request popup on initial load, I did a similar thing inside `ActivityLogContext`, it should be a simple copy-paste from there!

src/pages/[guild]/members.tsx
```diff
@@ -0,0 +1,227 @@
...
+            <OrderByColumn label="Join date" column={column} />
+          </HStack>
+        ),
+        cell: (info) => new Date(info.getValue()).toLocaleDateString(),
+      }),
+    ],
+    [hasHiddenRoles]
+  )
+
+  // TODO: keep row selection when the data changes. Right now we just reset the selection
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/members.tsx:136]: Should we do it in this PR, or it'll be an improvement in the future?

src/components/[guild]/Tabs/GuildTabs.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const showAccessHub =
+    (guildPlatforms?.some(
+      (guildPlatform) => guildPlatform.platformId === PlatformType.CONTRACT_CALL
+    ) ||
+      isMember ||
+      isAdmin) &&
+    !showOnboarding
+
+  const [eventsSeen, setEventsSeen] = useLocalStorage<boolean>("eventsSeen", false)
```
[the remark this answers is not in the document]
Comment (dovalid) [reply src/components/[guild]/Tabs/GuildTabs.tsx:31]: I've got to the decision to keep it this way until CRM is behind a feature flag for some guilds only, will change when everyone will see it

PR #838, opened 2023-06-19
V2 - User related endpoints

cs-balazs:
# V2 - User related endpoints
This PR contains necessary refactors for using the v2 endpoints for user related actions

Related BE PR: https://github.com/agoraxyz/guild-backend/pull/907


src/hooks/useKeyPair.ts
```diff
@@ -132,27 +135,59 @@ const setKeyPair = async ({
      * Ignoring this error is fine, since we are falling back to just storing it in
      * memory.
      */
-    await setKeyPairToIdb(userId, storedKeyPair).catch(() => {})
+    await setKeyPairToIdb(newUser.id, storedKeyPair).catch(() => {})
   }
 
-  await mutate(`/user/${account}`)
+  await mutate(
```
BrickheadJohnny, line 141: Do we need these awaits here, or the mutation will happen immediately since we're passing the new data in the second param? If we need to await each mutation, I wonder if we could use Promise.all, to do it all at once?

src/components/[guild]/hooks/useUser.ts
```diff
@@ -1,26 +1,71 @@
...
   const fetcherWithSign = useFetcherWithSign()
 
-  const { isLoading, data, mutate } = useSWRImmutable<User>(
-    account && ready && keyPair && isValid
-      ? [`/user/${account}`, { method: "GET", body: {} }]
+  const idToUse = userIdOrAddress ?? id
+
+  const { data, mutate, isLoading, error } = useSWRImmutable<User>(
+    idToUse && ready && keyPair && isValid
+      ? [`/v2/users/${idToUse}/profile`, { method: "GET", body: {} }]
```
Comment (dovalid) [inline src/components/[guild]/hooks/useUser.ts:37]: Isn't it the same as just passing the endpoint? Do we need a `{}` body for the authentication to work?

src/components/common/Layout/components/Account/components/AccountModal/components/SocialAccount.tsx
```diff
@@ -34,7 +34,7 @@ type Props = {
 const MotionHStack = motion(HStack)
 
 const SocialAccount = memo(({ type }: Props): JSX.Element => {
-  const { icon, name, colorScheme } = platforms[type]
+  const { icon, name, colorScheme } = platforms[type] ?? {}
```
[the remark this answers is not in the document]
cs-balazs: Nice catch! I believe this was only relevant while the endpoints were WIP! Removed it in https://github.com/agoraxyz/guild.xyz/pull/838/commits/4bdfd50d18e33c330c4e874b9b63cc9200308260!

src/components/[guild]/hooks/useUser.ts
Comment (cs-balazs) [reply src/components/[guild]/hooks/useUser.ts:37]: The original idea was that we use the endpoint as the key for unauthenticated requests, and this for authenticated ones. I believe this was done this way, so we can be sure that SWR will send the authenticated request since it has a different key (maybe with the current version it is different, but I think, that when optional auth was implemented, SWR didn't refetch on function change)
In this case, however, I think we can avoid this because the URLs are different, since the public one uses address, and the authenticated one uses userId. I have to test this out, but I think it should work

src/components/common/Layout/components/Account/components/AccountModal/hooks/useDisconnect.ts
```diff
@@ -1,37 +1,92 @@
...
+        method: "DELETE",
+        ...signedValidation,
+      }
+    )
+  }
+
+  return useSubmitWithSign<{ platformId: number }>(submit, {
+    onSuccess: ({ platformId }) => {
+      mutateUser(
+        (prev) => ({
```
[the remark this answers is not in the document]
cs-balazs: I think it would have been a great idea! I've removed the public one in https://github.com/agoraxyz/guild.xyz/pull/838/commits/bcdc51581c01f0f7c9d29d6f86d5126080bc6db5 as we discussed on Discord!

src/hooks/useKeyPair.ts
cs-balazs: Wow, nice! I never thought about this, but just checked, and it returns an instantly _fulfilled_ state Promise, so I think you are right, and the mutation itself is synchronous! Since we don't need the return values here, I've removed the awaits in https://github.com/agoraxyz/guild.xyz/pull/838/commits/a5e6f03522c34ee5e786535c044074ebf07c3be0!
I've tested it out, and seems to work as expected!
This is so cool, thanks for pointing it out!

PR #843, opened 2023-06-27
feat: NFT as reward

BrickheadJohnny:
This is a partial implementation of the NFT reward feature, it doesn't include the NFT creation flow - it'll be added in a separate PR.


src/components/[guild]/JoinModal/JoinModal.tsx
```diff
@@ -53,6 +53,10 @@ const JoinModal = ({ isOpen, onClose }: Props): JSX.Element => {
       const ConnectComponent = customJoinStep[platform]
       return <ConnectComponent key={platform} />
     }
+
+    // TODO: Temporary solution, we'll handle this automatically after the rewards/platforms refactor
```
Comment (dovalid) [inline src/components/[guild]/JoinModal/JoinModal.tsx:57]: This should be done on BE at the requiredPlatforms filter, as POAP is filtered there too

dovalid (approved):
Approved apart of the `Still WIP, no need to review` part in `useCollectNft.ts` and `TODO: call the v2 endpoint` in `[address].tsx`!

PR #851, opened 2023-07-13
feat: Guild tags

zgen-marci:


src/components/explorer/GuildCard.tsx
```diff
@@ -33,23 +38,38 @@ const GuildCard = ({ guildData }: Props): JSX.Element => (
...
+              fontSize="lg"
+              fontWeight="bold"
+              letterSpacing="wide"
+              maxW="full"
+              noOfLines={1}
+            >
+              {guildData.name}
+              {guildData.tags?.includes("VERIFIED") && (
+                <Tooltip label="Verified" hasArrow>
+                  <Circle
```
dovalid, line 60: The icon color is wrong in light mode. Also, should make a VerifiedIcon component and use that here and in the guild page for the title, so the design stays consistent

src/components/explorer/SearchBarFilters.tsx
```diff
@@ -0,0 +1,65 @@
...
+          bgColor={"transparent"}
+          onClick={() => onSelect(filter)}
+          position={"relative"}
+          zIndex={1}
+          {...(selected === filter && { _hover: { bg: "transparent" } })}
+        >
+          {filter.toLowerCase()}
+          {selected === filter ? (
+            <motion.div
+              className="underline"
```
dovalid, line 43: irrelevant className left here from example. Also, should only have one MotionBox or MotionCard element instead of having motion.div with the Card child. We can make any component a motion one like `const MotionCard = motion(Card)`. That way we can use Chakra props here too

dovalid (changes requested):

src/pages/[guild]/index.tsx
```diff
@@ -224,6 +226,9 @@ const GuildPage = (): JSX.Element => {
         backgroundImage={localBackgroundImage}
         action={isAdmin && <DynamicEditGuildButton />}
         backButton={{ href: "/explorer", text: "Go back to explorer" }}
+        titlePostfix={
+          tags && tags.includes("VERIFIED") && <VerifiedIcon iconSize={6} mt={-2} />
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/index.tsx:230]: I think we could use `tags?.includes("VERIFIED")` here, that would be a bit shorter

src/components/[guild]/EditGuild/components/TagManager.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const TagManager = ({ tags, setTags }: Props): JSX.Element => {
+  const { register, watch } = useForm<{ [k in GuildTags]: boolean }>({
+    defaultValues: {
+      VERIFIED: tags.includes("VERIFIED"),
+      FEATURED: tags.includes("FEATURED"),
+    },
+  })
+
+  useEffect(() => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/components/TagManager.tsx:20]: I think we could just use [useFormContext](https://react-hook-form.com/docs/useformcontext) here and put these tags directly into the edit guild form (so we wouldn't need this logic & the tags/setTags logic), then handle the API calls in handleSubmit (or, in other words, we could just put the post/delete requests from `useEditTags` to `useEditGuild`).

Also, it's pretty strange for me that we have separate endpoints just for the tags, but I guess that was a core decision?

src/components/explorer/GuildCard.tsx
```diff
@@ -33,23 +38,29 @@ const GuildCard = ({ guildData }: Props): JSX.Element => (
...
+              fontFamily="display"
+              fontSize="lg"
+              fontWeight="bold"
+              letterSpacing="wide"
+              maxW="full"
+              noOfLines={1}
+            >
+              {guildData.name}
+              {guildData.tags?.includes("VERIFIED") && (
+                <VerifiedIcon iconSize={5} position="absolute" top={1} right={-6} />
```
BrickheadJohnny, line 59: I think if we use an HStack instead of Box as a wrapper, we can remove the relative/absolute positioning and it'd look great without needing to position it this way

src/components/common/VerifiedIcon.tsx
```diff
@@ -0,0 +1,34 @@
+import { Circle, Container, Icon, SquareProps, Tooltip } from "@chakra-ui/react"
+import { Check, CircleWavy } from "phosphor-react"
+
+type Props = {
+  iconSize: number
+} & SquareProps
+
+const VerifiedIcon = ({ iconSize, ...chakraProps }: Props): JSX.Element => (
```
BrickheadJohnny, line 8: If you make some changes inside GuildCard (see my comment there), you can remove `...chakraProps` (`SquareProps`) here

src/components/common/VerifiedIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+          color={"blue.500"}
+          weight="fill"
+          position="absolute"
+        />
+        <Icon
+          as={Check}
+          boxSize={iconSize / 2}
+          color={"white"}
+          weight="bold"
+          position="absolute"
```
BrickheadJohnny, line 24: I'd refactor this component a bit. We could use a simple `Flex` or a `Center` component as a wrapper (we don't need the circular shape anymore), then remove the `Container`, and then remove the absolute position from the `Check` icon (it should be centered by default, since it's inside a flexbox)

BrickheadJohnny (changes requested):

src/types.ts
```diff
@@ -309,6 +310,9 @@ const supportedSocialLinks = [
 type SocialLinkKey = (typeof supportedSocialLinks)[number]
 type SocialLinks = Partial<Record<SocialLinkKey, string>>
 
+const guildTags = ["VERIFIED", "FEATURED"] as const
+export type GuildTags = (typeof guildTags)[number]
```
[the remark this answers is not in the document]
zgen-marci: There are other exports in the middle part of this file. Do I order them at the end of the file too?

src/components/[guild]/EditGuild/components/TagManager.tsx
Comment (zgen-marci) [reply src/components/[guild]/EditGuild/components/TagManager.tsx:20]: The Core decision was that use a separate tags endpoint to tend to RESTful. In the first version, I used the guild endpoint to edit the tags along with the other properties.

However, based on your suggestion, I'll try to make some changes here.

src/types.ts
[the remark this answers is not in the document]
BrickheadJohnny: Yup, feel free to move those to the end too!

src/components/[guild]/EditGuild/components/TagManager.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/EditGuild/components/TagManager.tsx:20]: All right, I think we could send these tag-related requests in `useEditGuild`, just like we did [in `useEditRole`](https://github.com/agoraxyz/guild.xyz/blob/main/src/components/%5Bguild%5D/RoleCard/components/EditRole/hooks/useEditRole.ts#L96-L108). What do you think?

src/components/[guild]/EditGuild/EditGuildDrawer.tsx
```diff
@@ -106,12 +100,19 @@ const EditGuildDrawer = ({
...
     socialLinks,
     featureFlags: isSuperAdmin ? featureFlags : undefined,
+    tags: savedTags,
   }
   const methods = useForm<GuildFormType>({
     mode: "all",
     defaultValues,
   })
 
+  const { onSubmit: onTagsSubmit } = useEditTags({
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/EditGuildDrawer.tsx:110]: We could improve the `useEditTags` hook a bit I think: you can easily retrieve `savedTags` and `id` (guild ID) inside the `useEditTags` hook (with the `useGuild` hook), so you don't need to pass these values as props, and then this hook could accept only 1 param (`currentTags`).

Also, you could just watch the `tags` like this:

```ts
const tags = useWatch({ control: methods.control, name: "tags" });
```

And then pass the watched value to the `useEditTags` hook:

```ts
const { onSubmit: onTagsSubmit } = useEditTags(tags);
```

BrickheadJohnny (approved):
🚀

dovalid (approved):

PR #852, opened 2023-07-13
Twitter v1 API support

cs-balazs:
# Twitter v1 API support

This PR adds support for the Twitter v1 OAuth API. The reason for supporting this older API is that it can return follower information unlike v2
As this is likely a temporary solution, some of the changes are somewhat hacky, the reason for that, is that our abstraction was made specifically for the OAuth 2.0 flow with authorization code, however, this one has some big differences:
- It needs an initial POST request to acquire a request token before the popup could be opened
  - This is the reason for the `oauthOptionsInitializer` parameter for `useOauthPopupWindow`
  - This POST request requires our app's API secret, which is why I added a serverless function
    - This function will be called, every time a Twitter v1 authentication happens, let me know if you think this will cost a lot
- It doesn't support the `state` query parameter
  - This is also the reason for most of the changes in `oauth.tsx`, like the one about the `BroadcastChannel`
- Added a hacky solution for displaying the correct requirement connect button, as we probably won't change the names of the requirements, so we just map them here. I'll need to make sure, which requirement types are affected exactly

There is one more inconvenience. If both of the Twitters are connected, they look the same in the account modal, I'm not sure how we should differentiate them, maybe with a tooltip, or a tag?


src/pages/api/twitter-request-token.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const CONSUMER_KEY = process.env.TWITTER_API_KEY
+const CONSUMER_SECRET = process.env.TWITTER_API_SECRET
+const SIGNATURE_METHOD = "HMAC-SHA1"
+const OAUTH_VERSION = "1.0"
+const DEFAULT_CALLBACK_URL = "https://guild.xyz/oauth"
+const BASE_URL = "https://api.twitter.com/oauth/request_token"
+const KEY = `${encodeURIComponent(CONSUMER_SECRET)}&`
+const RESPONSE_REGEX =
+  /oauth_token=(.*?)&oauth_token_secret=(.*?)&oauth_callback_confirmed=true/
```
Devidxyz, line 12: Is this a serverless endpoint? In that case this might be [ReDoS vulnerable](https://devina.io/redos-checker) too.

src/pages/api/twitter-request-token.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const handler: NextApiHandler = async (req, res) => {
+  if (req.method?.toLowerCase() !== "get") {
```
BrickheadJohnny, line 31: Really minor thing, but the we could use the `Allow` header here like we do on our other serverless endpoints ;)

```ts
if (req.method !== "GET") {
    res.setHeader("Allow", "GET")
    res.status(405).json({ error: `Method ${req.method} is not allowed` })
    return
}
```

src/pages/api/twitter-request-token.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const response = await fetch(url, {
+    method: "POST",
+    headers: {
+      Authorization: authHeaderValue,
+    },
+  })
+
+  if (!response.ok) {
+    res.status(400).json({ message: "Failed to generate Twitter request token" })
```
BrickheadJohnny, line 80: We could handle this error inside useOauthPopupWindow too, since right now I don't see an error message when the request fails, I can only see the Twitter oauth popup window, and an error inside that (it's because when we get this error when calling the serverless endpoint, we're sending an [object Object] in the oauth popup url as the oauth_token). ;)

src/pages/api/twitter-request-token.ts
```diff
@@ -0,0 +1,94 @@
...
+const CONSUMER_SECRET = process.env.TWITTER_API_SECRET
+const SIGNATURE_METHOD = "HMAC-SHA1"
+const OAUTH_VERSION = "1.0"
+const DEFAULT_CALLBACK_URL = "https://guild.xyz/oauth"
+const BASE_URL = "https://api.twitter.com/oauth/request_token"
+const KEY = `${encodeURIComponent(CONSUMER_SECRET)}&`
+const PARSE_FAIL_ERROR_MSG = "Unexpected data received from Twitter"
+
+// /oauth_token=(.*?)&oauth_token_secret=(.*?)&oauth_callback_confirmed=true/
+const parseV1Response = (response: string) => {
```
BrickheadJohnny, line 14: I think we could use `URLSearchParams` here! ;)
Haven't tested this code, but something like this maybe:
```ts
const params = new URLSearchParams(response)
const oauthToken = params.get("oauth_token")
const oauthTokenSecret = params.get("oauth_token_secret")
```

BrickheadJohnny (changes requested):

src/pages/api/twitter-request-token.ts
cs-balazs: I've made a similar change here, as I described in the core repo 👍 
https://github.com/agoraxyz/guild.xyz/pull/852/commits/fc3e224528edfa4f07ee222b19670b8476c63e80

src/pages/api/twitter-request-token.ts
cs-balazs: Wow, thank you! What an elegant way to handle this!
Made this change in https://github.com/agoraxyz/guild.xyz/pull/852/commits/616e4b3e51182f6ee2d2bf6f6678a42dc84b7711

src/pages/api/twitter-request-token.ts
cs-balazs: Great idea! Tested it out locally, and seems to work fine! Changed in https://github.com/agoraxyz/guild.xyz/pull/852/commits/cdf595693bedaf47b9454f24969cd94e888a69ec
Also made this change in the Twitter connector 👍

src/pages/api/twitter-request-token.ts
cs-balazs: Nice catch!
Added error handling in https://github.com/agoraxyz/guild.xyz/pull/852/commits/69c0970e92412e430d480daa3830a9d9ff27abcd
And added a PostHog event about this error in https://github.com/agoraxyz/guild.xyz/pull/852/commits/a6f8d7bd4ac0c8375371604aa07c094c77219a8d since we can't see these errors from the BE logs

PR #854, opened 2023-07-19
feat(processJoinPlatformError): pretty errors

BrickheadJohnny:
I've improved the `processConnectorError` util function, so it can handle this type of connector/runner errors too:

```
\"1\" runner error in resolvePlatformUser, params: {\"code\":\"...\",\"platformName\":\"DISCORD\",\"redirect_url\":\"https://guild.xyz/oauth\",\"scope\":\"guilds identify guilds.members.read\"}, error: \"failed to resolve user\"
```

### DIFF
--- a/src/components/[guild]/JoinModal/utils/processConnectorError.ts
+++ b/src/components/[guild]/JoinModal/utils/processConnectorError.ts
@@ -1,3 +1,5 @@
+import platforms from "platforms/platforms"
+import { PlatformType } from "types"
 import capitalize from "utils/capitalize"
 
 const processConnectorError = (error: string): string | undefined => {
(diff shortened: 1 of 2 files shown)


src/components/[guild]/JoinModal/utils/processConnectorError.ts
```diff
@@ -8,9 +10,25 @@ const processConnectorError = (error: string): string | undefined => {
...
+          PlatformType[parseInt(matchedPlatformId[0].replace('"', "").trim())]
+        ].name
+      : null
+    const cleanError = error.replaceAll("\\", "")
+    const matchedMsg = cleanError.match(/{"msg":"(.*)"}/gm)
+    const matchedError = cleanError.match(/error: "(.*)"/gm)
+
+    const parsedError = JSON.parse(
+      matchedMsg?.[0] ??
+        (matchedError?.[0]
```
Comment (BrickheadJohnny) [inline src/components/[guild]/JoinModal/utils/processConnectorError.ts:25]: Converting the error message to the `{ msg: "..." }` format, so we can use it easily inside `capitalize()`

Comment (cs-balazs) [reply src/components/[guild]/JoinModal/utils/processConnectorError.ts:25]: I'm not sure if it makes sense here or not, but without the `g` flag we could just access the captured group, and do `matchedMsg ?? matchedError` I think, what do you think?

I mean something like this:
```ts
const matchedMsg = cleanError.match(/{"msg":"(.*?)"}/m)?.[1]
```

cs-balazs (approved):
Nice handling of these errors!
I think something similar will need to be implemented on the BE side as well to get cleaner log messages 🤔 
I've left 2 small questions/suggestions, but I don't think they affect functionality!

src/components/[guild]/JoinModal/utils/processConnectorError.ts
Comment (BrickheadJohnny) [reply src/components/[guild]/JoinModal/utils/processConnectorError.ts]: Oh, really nice catch, thanks, updated it in 2d5c527bb2ff1ca85e0a2e314a7cfaea3e8fe150
The `.match()` method actually always returns an array, but I removed the `g` flag and desctructured the array, so the code is a bit cleaner now.

PR #858, opened 2023-07-26
OAuth improvements

cs-balazs:
# OAuth improvements

This PR aims to both solve the hook call order problem of the current OAuth solution and to generalize the whole authentication flow with the `platforms` object

Sidequest: Experiment with TG, maybe there is a workaround to use that as well


src/platforms/platforms.ts
```diff
@@ -65,18 +109,55 @@ const platforms: Record<PlatformName, PlatformData> = {
     gatedEntity: "repo",
     cardPropsHook: useGithubCardProps,
     cardMenuComponent: GithubCardMenu,
+
+    oauth: {
+      url: "https://github.com/login/oauth/authorize",
+      params: {
+        client_id: process.env.NEXT_PUBLIC_GITHUB_CLIENT_ID,
+        scope: {
+          creation: "repo,read:user",
```
BrickheadJohnny, line 118: I believe we actually need this scope for some of the requirements too, so not sure what could we do regarding this. If we'd request this scope from everyone, then the users wouldn't be too happy, so maybe we should send the "creation" scope in the reconnect button? (because we only show a reconnect button for the users who haven't authenticated with the required scope via GitHub - that's why we added that ternary a couple of months ago in `ConnectRequirementPlatformButton`)

src/platforms/platforms.ts
```diff
@@ -36,8 +45,16 @@ type PlatformData = {
   }
   cardSettingsComponent?: () => JSX.Element
   cardMenuComponent?: (props) => JSX.Element
-  cardWarningComponent?: (props) => JSX.Element
+  cardWarningComponent?: (props) => ReactNode
```
BrickheadJohnny, line 48: I wonder if this change is needed? 👀

src/components/[guild]/Requirements/components/ConnectRequirementPlatformButton.tsx
```diff
@@ -55,13 +55,10 @@ const ConnectRequirementPlatformButton = (props: ButtonProps) => {
     (err) => err.requirementId === id && err.errorType === "PLATFORM_CONNECT_INVALID"
   )
 
-  const scope = platform === "GITHUB" ? "repo,read:user" : undefined
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/components/ConnectRequirementPlatformButton.tsx:58]: Added a comment regarding this in `platforms.ts` 😉

src/platforms/platforms.ts
BrickheadJohnny: Nvm, we talked about it on Discord, I guess we can revert this change, haha!

src/platforms/platforms.ts
cs-balazs: Thanks for pointing this out! This is an interesting scenario
I think I might rename the scope levels to _basic_ and _extended_ or something similar
I'm not sure whether we should first request _basic_ level and then request _extended_ when needed (on reconnect, as you recommended), or request for all the scopes, we will need 🤔 

It looks to me, that currently on the account modal, we request _basic_ scopes, but on a requirement card we request _extended_ scopes (for both reconnect and connect)

src/platforms/platforms.ts
cs-balazs: Jeje, thanks, reverted in https://github.com/agoraxyz/guild.xyz/pull/858/commits/41b2888f87fa66eb9e03b73c3107db4287e09ca2

src/platforms/platforms.ts
cs-balazs: For now I changed it to request _creation_ scopes on reauth in https://github.com/agoraxyz/guild.xyz/pull/858/commits/506ce2bdc9627878c31096607d02bdd756030770

BrickheadJohnny: Great, I think it's good this way for now!

PR #864, opened 2023-08-04
Fix non-discord templated guild creation

cs-balazs:
# Fix non-discord templated guild creation

~~Telegram guild creation currently fails when using the growth template, because it adds the Telegram reward to both roles by default. This PR aims to fix this issue by only adding the Telegram reward to one of the roles~~

I realized this is a problem with GitHub and Google as well, not just Telegram, so I've changed it to only include the second reward if the platform is Discord


dovalid (approved):

MTtalib:
Nice

PR #865, opened 2023-08-07
platforms.ts & add reward modal refactor

BrickheadJohnny:
While I was developing the NFT as reward creation flow, I noticed that we don't handle the usage restrictions for each platform properly and I also realized that we'd need a way to easily create a custom reward creation flow, where we don't follow the "add reward then pick role(s) for it" flow, so this PR aims to solve these 2 issues.

**Usage restrictions**
- added a `PlatformUsageRestrictions` enum with `SINGLE_ROLE`, `MULTIPLE_ROLES`, and `NOT_APPLICABLE` values
- added a `usageRestrictions` property to each platform in `platforms.ts`
- checking this prop where we need to display different things based on the usage restrictions (e.g. Create guild page & Add Reward modal)
- added a generalized `usePlatformUsageInfo` hook, which we can use to determine if a platform can be added to a new role or not (without checking its usageRestriction prop manually) - we're using this hook both on the create guild page (or to be precise, inside the platform selector cards: DCServerCard, RepoCard, GoogleDocCard) and inside the add reward modal
- deleted the `useGuildByPlatformId` hook, its functionality is now only used in this new `usePlatformUsageInfo` hook

**Add reward modal refactor**
- created a `RoleSelector` component, it can be used both as a radio group (where you can select only 1 role) and as a checkbox group (for selecting multiple roles)
- refactored the add reward modal, so we can either use a custom reward creation component or use the default one (where the user can add a reward and then select a role or multiple roles for it) 
- added a `AddPlatformPanel` prop for platforms in `platforms.ts` - it accepts a dynamically loaded component and is used for adding rewards, it'll be shown before the role selector / role creation step in the add reward flow


src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/SelectExistingPlatform.tsx
```diff
@@ -48,7 +46,8 @@ const SelectExistingPlatform = ({ onClose }) => {
             "application/vnd.google-apps.form"
 
           const isAddButtonDisabled =
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/SelectExistingPlatform.tsx:48]: Yes, we could use the new `usePlatformUsageInfo` hook here, but this solution is actually simpler in this specific case, so I left it as is.

src/platforms/hooks/usePlatformUsageInfo.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const usePlatformUsageInfo = (
+  platform: PlatformName,
+  platformId: string
+): { isAlreadyInUse: boolean; guildUrlName?: string; isValidating: boolean } => {
+  const { usageRestriction } = platforms[platform]
+
+  const { id } = useGuild()
+
+  // Not sure if it'll be a problem (API performance wise) that we call this endpoint here, but
```
BrickheadJohnny, line 14: Maybe Balazs can provide some info regarding this, then we can remove this comment ;)

cs-balazs: I think the fact that we request them one by one shouldn't be a big problem, I think we did it previously as well. Two think I'm not sure about:
- Do we need it to be mutable?
- We should avoid resending requests that resulted in 404. I could think of these two possible solutions:
  1) Map 404 errors to successes in this SWR call's fetcher. A bit messy, since we need to pass the fetcher manually
  2) Disable `shouldRetryOnError`
  - In the latter case, we might run into cases where a revalidation might be necessary, therefore the other solution might be better. That way we only "ignore" the 404 responses, which is fine, since from a 404 response, we already know all the information we needed. Let me know which approach you prefer, or if you have an alternative!

cs-balazs (commented):
Fantastic refactor! 🚀 Left some minor suggestions and some questions!

src/platforms/platforms.ts
```diff
@@ -38,7 +64,8 @@ type PlatformData = {
   cardMenuComponent?: (props) => JSX.Element
   cardWarningComponent?: (props) => JSX.Element
   cardButton?: (props) => JSX.Element
-}
+  usageRestriction: PlatformUsageRestrictions
```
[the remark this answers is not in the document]
BrickheadJohnny: Changed it in 1950aaffbfeb3f570e26c979d4f502b41615506e!

src/components/[guild]/CreatePoap/AddPoapModalContent.tsx
```diff
@@ -0,0 +1,90 @@
...
+  goBack: () => void
+  onSuccess: () => void
+}
+
+const AddPoapModalContent = ({
+  modalRef,
+  goBack,
+  onSuccess,
+}: Props): JSX.Element => {
+  const [step, setStep] = useState("home")
```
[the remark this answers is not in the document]
Comment (BrickheadJohnny) [reply src/components/[guild]/CreatePoap/AddPoapModalContent.tsx]: Oh yep, thanks for pointing this out, I just copy-pasted the old POAP creation component and missed this! Added types in 5cf7630e44fe3f9cdabf13e5c38d9cbc48ae25d8

src/platforms/hooks/usePlatformUsageInfo.ts
BrickheadJohnny: Thanks for the ideas! I made it immutable & decided to just disable `shouldRetryOnError` for now, but we can talk more about it on DC!

PR #873, opened 2023-08-23
Ethereum attestation requirement

cs-balazs:
# Ethereum attestation requirement

- Adds a new Ethereum attestation requirement
  - I've created a test attestation, which we can use to test different key-value configurations (on sepolia): https://sepolia.easscan.org/schema/view/0x0dc7f776c8ff53eeec68f4bf555751684b4e98146af45b645d1d7251f4af3eec
- The UX could be improved a lot, especially for array keys, but with @marclllaks we thought, that it is important to have the requirement up and running for those who need it, and we could improve on it in a follow-up PR
  - We could fetch the possible keys and their types from easscan, and have the user pick from them
  - And we could have a special input field, like the one for NFT traits for arrays
- `BlockExplorerUrl` has been changed to accept a `label` and a `path` parameter. This was needed to display this component with "Attester" or "Reciever" labels, and to fix the path to "address"


src/requirements/EthereumAttestation/EthereumAttestationRequirement.tsx
```diff
@@ -0,0 +1,94 @@
...
+      }
+      footer={
+        <HStack spacing={4}>
+          <RequirementLinkButton
+            href={`${EAS_SCAN_BASE[requirement?.chain ?? "ETHEREUM"]}/${
+              requirement?.data?.schemaId
+            }`}
+            imageUrl="/requirementLogos/eas.png"
+          >
+            {/* View schema on EAS scan */}
```
BrickheadJohnny, line 44: I think you can remove these comments if you decide which copy should we use ;)

src/requirements/EthereumAttestation/components/EthereumAttestation.tsx
```diff
@@ -0,0 +1,119 @@
...
+import FormErrorMessage from "components/common/FormErrorMessage"
+import { Chain } from "connectors"
+import { useFormContext, useWatch } from "react-hook-form"
+import { RequirementFormProps } from "requirements"
+import ChainPicker from "requirements/common/ChainPicker"
+import parseFromObject from "utils/parseFromObject"
+import { EAS_SCAN_BASE } from "../EthereumAttestationRequirement"
+
+const HEX_STRING_REGEX = /^0x[A-F0-9]+$/i
+const ADDRESS_REGEX = /^0x[A-F0-9]{40}$/i
```
BrickheadJohnny, line 11: nit: we export this regex from another file, we can use it from there, so we have a "single source of truth" (I know it's located in a guild checkout related file, maybe we'll move it to a more appropriate file in the future)

src/requirements/EthereumAttestation/EthereumAttestationRequirement.tsx
```diff
@@ -0,0 +1,94 @@
...
+            label={requirement.type === "EAS_ATTEST" ? "Recipient" : "Attester"}
+          />
+        </HStack>
+      }
+      {...props}
+    >
+      {requirement.type === "EAS_ATTEST" ? (
+        <>
+          Attest{" "}
+          <DataBlock>
```
BrickheadJohnny, line 62: We could use the `DataBlockWithCopy` component for hex strings, so the users can easily copy it if needed, maybe that would improve the UX a bit.

BrickheadJohnny (approved):
Added a couple of suggestions, but lgtm overall, great work! We can implement the UX improvements which you mentioned in the PR description later! 👍

src/requirements/EthereumAttestation/EthereumAttestationRequirement.tsx
cs-balazs: I think the current version displaying only "Attester", "Recipient" or "Schema" is better, removed these comments in https://github.com/agoraxyz/guild.xyz/pull/873/commits/1a88792d9308c6a9971859a949ec5458a2558e51 👍

src/requirements/EthereumAttestation/components/EthereumAttestation.tsx
cs-balazs: Thanks! Definitely better to have this defined only once! Made the change in https://github.com/agoraxyz/guild.xyz/pull/873/commits/0c49d0a10ae6d877b4a03a8e00dacb6b5e4a23c5 !

src/requirements/EthereumAttestation/EthereumAttestationRequirement.tsx
cs-balazs: Nice! Thanks for mentioning this! At one point I was even thinking about implementing something like this, didn't notice, that it already has a nice component (then instead of implementing this, I just put the links in the footer 😄 ). Made the change in https://github.com/agoraxyz/guild.xyz/pull/873/commits/33278e1323f8aec7494cb2c008d0c969f0459f3b
Added this for the attester/recipient, and for the schema

PR #875, opened 2023-08-24
ERC-1155 UI changes

cs-balazs:
# ERC-1155 UI changes

This PR includes some small UI changes according to feedback here: https://discord.com/channels/697041998728659035/1143520047344652339/1144229221737189496

- `NftRequirement`: display the id, which the requirement is checking for, and use "a(n)" instead of "the" for `ERC1155`
- `BlockExplorerUrl`: Checks if the current requirement is either `ERC721` or `ERC1155`, and has a `data.id` field and if it does, use a link directly to that id


BrickheadJohnny (approved):
🚀

PR #878, opened 2023-08-30
fix: drawer and modal animation lag

zgen-marci:
Symptom:
Drawers and Modals open laggy, poor user experience.

Hypothesis:
The useKeyPair hook needs lots of calculations during the animation.

Cure:
Move the heavy logic inside the useKeyPair hook into an app-level KeyPair context to save some recalculation.


src/components/_app/useKeyPairContext.tsx
```diff
@@ -431,5 +702,7 @@ const useKeyPair = () => {
   }
 }
 
-export { deleteKeyPairFromIdb, getKeyPairFromIdb, setKeyPairToIdb }
+const useKeyPair = () => useContext(KeyPairContext)
+
+export { KeyPairProvider, deleteKeyPairFromIdb, getKeyPairFromIdb }
 export default useKeyPair
```
BrickheadJohnny, line 464: We can remove the default export and just use named export, just like we do it in every other `...Context` components:

```ts
export { KeyPairProvider, useKeyPair, deleteKeyPairFromIdb, getKeyPairFromIdb }
```

src/components/_app/useKeyPairContext.tsx
```diff
@@ -180,7 +180,278 @@ const setKeyPair = async ({
...
+  set: any
+}>({
+  ready: false,
+  pubKey: undefined,
+  keyPair: undefined,
+  isValid: false,
+  set: () => {},
+})
+
+const KeyPairProvider = ({ children }: PropsWithChildren<unknown>): JSX.Element => {
```
BrickheadJohnny, line 204: The file name could be `KeyPairProvider` so we stay consistent ;) *(at least under this directory, haha)*

src/components/_app/PostHogProvider.tsx
```diff
@@ -34,7 +35,8 @@ const PostHogContext = createContext<{
 const CustomPostHogProvider = ({
   children,
 }: PropsWithChildren<unknown>): JSX.Element => {
-  const { id, addresses } = useUser()
+  const { account } = useWeb3React()
```
BrickheadJohnny, line 38: `useUserPublic` doesn't contain the user address, so we decided we'll just use the address from web3-react here

BrickheadJohnny (commented):
Left a couple of comments, please check them, but overall, I think this is a great improvement! 🎉

src/components/_app/useKeyPairContext.tsx
zgen-marci: The other files ended with the name Provider export only the Context and the Provider itself, but this one exports some other tools. What do you think about moving the IDB-related util function into a different file in a global util dictionary?

BrickheadJohnny: Good point, but since both `getKeyPairFromIdb` and `setKeyPairToIdb` are keypair-related functions, I think we can leave them in this file and just export them from here ;)

BrickheadJohnny (approved):
Great work, thank you! 🚀

dovalid (approved):

PR #880, opened 2023-09-01
hotfix: optional auth SWR error

cs-balazs:
# hotfix: optional auth SWR error

Due to https://github.com/agoraxyz/guild.xyz/commit/add059fe49efd926f106ed099a090abffb5b99bb, guild data was `undefined` for the duration of the authenticated request. This PR fixes this by binding the unauthenticated request's `shouldFetch` to `authenticatedResponse.data`, so it only stops fetching once we have authenticated data


thaerabuissa:
HEJ

PR #887, opened 2023-09-11
fix: put `addressLinkParams` in its own context

BrickheadJohnny:
### Describe the bug

Users currently can't link new addresses to their current profile,  because it seems like properties inside `addressLinkParams` are not defined. It happens because we store these values inside `Web3ConnectionManager`, but we use them in `KeyPairProvider`, which is **not** wrapped inside the connection manager provider.

### Steps to reproduce

1. Connect your wallet to Guild.xyz
2. Open the account modal, and click on "+ Link address"
3. Try linking another wallet address
4. It just won't work (we don't show an error toast, or anything else)

### How this PR solves the issue

I extracted the `addressLinkParams` state in its own context and moved it above `KeyPairProvider`.

### DIFF
--- a/src/components/_app/AddressLinkProvider.tsx
+++ b/src/components/_app/AddressLinkProvider.tsx
@@ -1,0 +1,31 @@
+import {
+  createContext,
+  Dispatch,
+  PropsWithChildren,
+  SetStateAction,
+  useContext,
+  useState,
+} from "react"
+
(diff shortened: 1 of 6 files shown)


cs-balazs (approved):
Nice! Managed to reproduce the bug in prod, and this branch worked! 🚀

PR #888, opened 2023-09-12
fix: add userId to generatedKeyPair SWR key

BrickheadJohnny:
The keypair wasn't invalidated when the user switched to another wallet, so we sent an invalid public key to our API. I added the userId to the SWR key in this PR, so it'll be regenerated on wallet change.


cs-balazs (approved):
🚀

PR #890, opened 2023-09-13
feat: display correlation id on error toasts

BrickheadJohnny:
## Feature

The main idea was to catch Guild API errors in our custom fetcher, and return them in a uniform format:

```ts
{
  error: string,
  correlationId: string
}
```

This way we can display correlation IDs on error toasts.

### Notes
- I'm not sure why, but `useClipboard` didn't work inside `useShowErrorToast`, so I simply used `navigator?clipboard.writeText()`
- I also needed to modify the complex logic inside `useEditGuild` & `useEditRole`, I think I didn't break anything, but please double check it ;)
- Also, if you have a cleaner solution/idea for this, please lmk, I can rework it if needed


src/components/[guild]/EditGuild/hooks/useEditGuild.ts
```diff
@@ -250,31 +261,55 @@ const useEditGuild = ({ onSuccess, guildId }: Props = {}) => {
...
-          showErrorToast("Failed to delete some feature flags")
+          showErrorToast({
+            error: "Failed to delete some feature flags",
+            correlationId: featureFlags.deletions.correlationId,
+          })
         }
 
-        if (guildUpdateResult === null) {
-          showErrorToast("Failed to update guild data")
+        if (guildUpdateResult.error) {
```
Comment (cs-balazs) [inline src/components/[guild]/EditGuild/hooks/useEditGuild.ts:308]: I think we need a `?` here, because it can be `undefined`
```suggestion
        if (guildUpdateResult?.error) {
```

src/components/[guild]/EditGuild/hooks/useEditGuild.ts
```diff
@@ -250,31 +261,55 @@ const useEditGuild = ({ onSuccess, guildId }: Props = {}) => {
         })
       } else {
         if (admin.creations.failedCount > 0) {
-          showErrorToast("Failed to create some admins")
+          showErrorToast({
```
Comment (cs-balazs) [inline src/components/[guild]/EditGuild/hooks/useEditGuild.ts:264]: I know it is not really the scope, but I think in the bigger if condition  above, we need a `(!guildUpdateResult || (!!guildUpdateResult && !guildUpdateResult.error))` at the end instead of just `guildUpdateResult`
Because if a guild update doesn't happen, and other requests succeed, then `guildUpdateResult` is undefined, and we won't show a toast

cs-balazs (commented):
Nice changes! I tried to think of an alternative solution, but all I could think of is a separate `Error` class, like `GuildAPICallError` which we throw from `fetcher`, but I don't think it has any benefits (maybe some checks would look nicer, like `error instanceof GuildAPICallError`), just mentioning it :ayy:

src/components/[guild]/EditGuild/hooks/useEditGuild.ts
Comment (BrickheadJohnny) [reply src/components/[guild]/EditGuild/hooks/useEditGuild.ts:264]: Oh right, nice catch! I also noticed an issue in `useEditRole`, and fixed that in my latest commit too!

PR #891, opened 2023-09-14
PulseMarker and tryToParseJSON refactor

BrickheadJohnny:
## Bugs

- The `tryToParseJSON` function returned the default (`false`) value, when it should not have, so I refactored & simplified it a bit
- If we changed the `hidden` prop of the `PulseMarker` component based on a value which we fetched from localStorage (only on the client), we got a hydration error. Now we import the marker/animation dynamically & only on the client, so this won't be an issue,


cs-balazs (approved):
Lgtm 🚀

PR #895, opened 2023-09-18
fix: add back to explorer and setting icon to events page, sort event…

zgen-marci:
https://linear.app/guildxyz/issue/GUILD-1174/discord-event-feedbacks


BrickheadJohnny (approved):
Requested a minor change, but otherwise looks good to me, nice changes!

PR #899, opened 2023-09-20
feat: add guild url to posthog metadata on events tab click

zgen-marci:


src/pages/[guild]/activity.tsx
```diff
@@ -55,7 +55,10 @@ const ActivityLog = (): JSX.Element => {
             href={`/${urlName}/events`}
             onClick={() => {
               setEventsSeen(true)
-              captureEvent("Click on events tab", { from: "activity log" })
+              captureEvent("Click on events tab", {
+                from: "activity log",
+                guildUrl: urlName,
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/activity.tsx:60]: Please use `guild: urlName`, so we can filter this event the same way we can filter other events ;)

PR #905, opened 2023-09-22
feat: event source editor & new sources

zgen-marci:
https://linear.app/guildxyz/issue/GUILD-1100/events-admin-section


src/components/[guild]/EditGuild/EditGuildDrawer.tsx
```diff
@@ -97,6 +99,11 @@ const EditGuildDrawer = ({
     socialLinks,
     featureFlags: isSuperAdmin ? featureFlags : undefined,
     tags: savedTags,
+    eventSources: {
+      EVENTBRITE: eventSources.EVENTBRITE,
+      LINK3: "",
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/EditGuildDrawer.tsx:104]: Is this intended? Or should it be `eventSources.LINK3`?
Edit: now I see, so we don't support LINK3 yet?

src/components/[guild]/EditGuild/components/Events/EventInput.tsx
```diff
@@ -0,0 +1,58 @@
...
+  LINK3: "Link3",
+  LUMA: "lu.ma",
+  DISCORD: "Diiscord",
+}
+
+const logos: Record<EventSourcesKey, string> = {
+  EVENTBRITE: "/platforms/eventbrite.png",
+  LINK3: "/platforms/link3.png",
+  LUMA: "/platforms/luma.png",
+  DISCORD: "platforms/discord.png",
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/components/Events/EventInput.tsx:28]: I think we need a `/` here?

src/components/[guild]/EditGuild/components/Events/EventInput.tsx
```diff
@@ -0,0 +1,58 @@
...
+  Input,
+  InputGroup,
+  InputLeftElement,
+  InputRightElement,
+} from "@chakra-ui/react"
+import { useFormContext } from "react-hook-form"
+import { EventSourcesKey, GuildFormType } from "types"
+
+type Props = {
+  name: EventSourcesKey
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/components/Events/EventInput.tsx:14]: I'd call this prop `eventSource`, so we know exactly what's this (it wasn't clear when I first saw this component)

src/components/[guild]/Events/DiscordEventCard/components/JoinDiscordEventButton.tsx
```diff
@@ -2,31 +2,49 @@ import { ButtonProps } from "@chakra-ui/react"
...
   const { captureEvent } = usePostHogContext()
 
   return (
     <Button
       as="a"
-      href={`https://discord.com/events/${guildId}/${eventId}`}
+      href={url}
       target="_blank"
-      colorScheme="indigo"
+      bg={eventButtonColor[eventType]}
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Events/DiscordEventCard/components/JoinDiscordEventButton.tsx:44]: I mentioned this on Discord, but we could use `useColorPalette` hook here to generate the necessary swatches so the button would look great in hover/focus/active states too!

src/components/[guild]/EditGuild/components/Events/Events.tsx
```diff
@@ -0,0 +1,23 @@
...
+  <>
+    <Text colorScheme="gray">
+      Guild can auto-import your events from different platforms to show them in one
+      place.
+    </Text>
+    <SimpleGrid columns={2} gap={3}>
+      <EventInput name={"EVENTBRITE"} />
+      <EventInput name={"LUMA"} />
+
+      {/*supportedEventSources
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/components/Events/Events.tsx:14]: You can remove this comment ;) And then add back when we'll actually have more event sources

src/pages/[guild]/events.tsx
```diff
@@ -51,14 +52,15 @@ const GuildEvents = (): JSX.Element => {
 
   const discordGuildPlatform = guildPlatforms?.find(
     (platform) => platform.platformId === PlatformType.DISCORD
-  )
+  )?.platformGuildId
 
-  const { data, isLoading, error } = useDiscordEvents(
-    discordGuildPlatform?.platformGuildId
+  const { data, isLoading, error, serverError } = useGuildEvents(
+    discordGuildPlatform,
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/events.tsx:58]: Consider using the `useGuild` hook inside `useGuildEvents` - that way you wouldn't need to pass these props

src/pages/[guild]/events.tsx
```diff
@@ -133,39 +135,40 @@ const GuildEvents = (): JSX.Element => {
...
-
-      {!isLoading && !!error && (
-        <FallbackFrame
-          icon={WarningOctagon}
-          text="Something went wrong, couldn't load events."
-        />
-      )}
-
-      {!isLoading && !error && data?.length === 0 && (
+      {isLoading ||
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/events.tsx:138]: I think this is why there's no loading state on initial load? 👀 I think it should be `(isLoading || data === undefined) && ...`?

src/hooks/useGuildEvents.ts
```diff
@@ -0,0 +1,99 @@
...
+  entityId?: string
+  userCount?: number
+  creator?: number
+  entityMetadata?: {
+    location: string
+  }
+  image?: string
+}
+
+const DiscordEventToGuildEvent = (
```
BrickheadJohnny, line 37: let's use camelCase for function names (`discordEventToGuildEvent`), and I think we could rename the `guildId` param to `serverId`

src/hooks/useGuildEvents.ts
```diff
@@ -0,0 +1,99 @@
...
+  const swrResponseDiscord = useSWRImmutable<
+    OneOf<{ events: DiscordEvent[] }, { error: any }>
+  >(platformGuildId ? `/discord/events/${platformGuildId}` : null)
+
+  const data = []
+
+  if (swrResponseEvents.data?.events) data.push(...swrResponseEvents.data?.events)
+  if (swrResponseDiscord.data?.events)
+    data.push(
+      ...swrResponseDiscord.data?.events.map((event) =>
```
BrickheadJohnny, line 69: I think both of these `push` calls will fail if events is undefined, so we could define a fallback value, like this:
```ts
data.push(...(swrResponseEvents.data?.events ?? []))
```

src/pages/[guild]/events.tsx
```diff
@@ -133,39 +135,40 @@ const GuildEvents = (): JSX.Element => {
...
-      {!isLoading && !error && data?.length > 0 && (
+      {!isLoading && data?.length > 0 && (
         <VStack gap={4}>
           {data.sort(sortEventByStartDate).map((event) => (
-            <DiscordEventCard key={event.id} event={event} guildId={guildId} />
+            <DiscordEventCard key={event.title} event={event} guildId={guildId} />
           ))}
         </VStack>
       )}
+      {isAdmin && !isLoading && error.length ? (
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/events.tsx:156]: nit: we could construct 1 error message inside `useGuildEvents` and just display that here, so we wouldn't need to do all this logic
(if this is blocking, let's refactor this later in a separate pr)

BrickheadJohnny (commented):
Really good work overall, left a couple of comments!

src/hooks/useGuildEvents.ts
BrickheadJohnny: Nevermind, I see the if statement now

BrickheadJohnny (approved):

PR #915, opened 2023-09-30
feat(ContractStateForm): define values manually

BrickheadJohnny:
This PR adds support for the `CONTRACT` requirement type on all supported networks by allowing the users to fill out the form manually instead of selecting values from select inputs. The original functionality still works if we can fetch the contract ABI from the API.

### DIFF
--- a/src/requirements/ContractState/ContractStateForm.tsx
+++ b/src/requirements/ContractState/ContractStateForm.tsx
@@ -1,30 +1,44 @@
 import {
   Divider,
   FormControl,
+  FormHelperText,
   FormLabel,
   HStack,
+  IconButton,
   Input,
+  InputGroup,
+  InputLeftAddon,
+  InputRightElement,
+  NumberDecrementStepper,
+  NumberIncrementStepper,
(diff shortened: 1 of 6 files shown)


src/requirements/ContractState/ContractStateForm.tsx
```diff
@@ -64,11 +83,7 @@ const ContractStateForm = ({ baseFieldPath }: RequirementFormProps) => {
   )
 
   const methodData = useMemo(
-    () =>
-      abi &&
-      method &&
-      (abi.find((m) => m.name === method.split("(")[0]) ||
-        setValue(`${baseFieldPath}.data.id`, null)),
```
BrickheadJohnny, line 71: I wasn't sure why we needed this, @dovalid I think you added this line, do you remember by any chance? 👀

dovalid: Not exactly, it should have been some edge case resetting mechanism, it should be safe to remove

dovalid (approved):

PR #923, opened 2023-10-04
feat: `useSimpleGuild` & `useRole` hooks

BrickheadJohnny:
Added these 2 new hooks, so we don't fetch the guild data from the `/v2/guilds/:id/guild-page` endpoint inside `GuildRequirement`.

### DIFF
--- a/src/components/[guild]/Requirements/components/GuildCheckout/MintGuildPin/components/ActivateGuildPinForm.tsx
+++ b/src/components/[guild]/Requirements/components/GuildCheckout/MintGuildPin/components/ActivateGuildPinForm.tsx
@@ -7,6 +7,7 @@ import { Chain } from "connectors"
 import useToast from "hooks/useToast"
 import { FormProvider, useController, useForm, useWatch } from "react-hook-form"
 import ChainPicker from "requirements/common/ChainPicker"
+import { Visibility } from "types"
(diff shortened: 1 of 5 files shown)


cs-balazs (approved):
Great changes! 🚀 
I'm wondering if it would be worth it to include a custom fetcher function in these hooks that first checks if the `/guild-page` data is already in SWR state, and if it is, it uses the data from there. We might avoid some requests for cases when a requirement references a role from its own guild 🤔 
But these requests are very lightweight, so it is not important at all, just sharing my thoughts :ayy:

BrickheadJohnny:
> Great changes! 🚀 I'm wondering if it would be worth it to include a custom fetcher function in these hooks that first checks if the `/guild-page` data is already in SWR state, and if it is, it uses the data from there. We might avoid some requests for cases when a requirement references a role from its own guild 🤔 But these requests are very lightweight, so it is not important at all, just sharing my thoughts :ayy:

Ooh, that's a great idea! I made a Linear issue about it ([GUILD-1397](https://linear.app/guildxyz/issue/GUILD-1397/usesimpleguild-check-swr-cache-before-fetching)), so we don't forget it, but for now I'll merge it as is (the current solution is still better than always fetching the whole guild-page data hehe). Thanks for the review!

PR #925, opened 2023-10-05
Fix/guild events save issue

zgen-marci:
Issue: The user can't save the guild after changing the event link. The reason was the guild form sent an empty string in case some event link wasn't provided. The backend handled the empty string as a wrong link format and gave an error on save.

Solution: The front end won't register links until the user manually adds to the event source list. By the way, this was a feature request partially. 

See details here:
https://linear.app/guildxyz/issue/GUILD-1348/ultimate-event-admin-ui

### DIFF
--- a/src/components/[guild]/EditGuild/components/Events/EventInput.tsx
+++ b/src/components/[guild]/EditGuild/components/Events/EventInput.tsx
@@ -7,7 +7,7 @@ import {
   InputLeftElement,
   InputRightElement,
 } from "@chakra-ui/react"
-import { useFormContext } from "react-hook-form"
+import { useFormContext, useWatch } from "react-hook-form"
 import { EventSourcesKey, GuildFormType } from "types"
 
 type Props = {
@@ -30,6 +30,13 @@ const logos: Record<EventSourcesKey, string> = {
 
 const EventInput = ({ eventSource }: Props) => {
   const { register, setValue } = useFormContext<GuildFormType>()
+  const link = useWatch({
+    name: `eventSources.${eventSource}`,
+  })
+
+  console.log("xy", link)
+
+  if (!link && link !== "") return null
 
   return (
     <InputGroup size={"lg"}>
@@ -55,4 +62,5 @@ const EventInput = ({ eventSource }: Props) => {
   )
 }
 
+export { logos, placeholders }
 export default EventInput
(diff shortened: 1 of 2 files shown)


src/components/[guild]/EditGuild/components/Events/EventInput.tsx
```diff
@@ -29,34 +31,46 @@ const logos: Record<EventSourcesKey, string> = {
...
+    register,
+    setValue,
+    formState: { errors },
+  } = useFormContext<GuildFormType>()
+
   const link = useWatch({
     name: `eventSources.${eventSource}`,
   })
 
   if (!link && link !== "") return null
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/components/Events/EventInput.tsx:44]: I know this isn't a new line, but why do we need this exactly?

src/components/[guild]/EditGuild/components/Events/EventInput.tsx
```diff
@@ -29,34 +31,46 @@ const logos: Record<EventSourcesKey, string> = {
...
-      <InputRightElement>
-        <CloseButton
-          aria-label="Remove link"
-          size="sm"
-          rounded="full"
-          onClick={() => setValue(`eventSources.${eventSource}`, undefined)}
+    <FormControl isInvalid={!!errors?.eventSources?.[eventSource]} isRequired>
+      <InputGroup size={"lg"}>
+        <InputLeftElement>
+          <Circle bgColor={"gray.900"} size={5}>
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/components/Events/EventInput.tsx:50]: I think we'll always have an event source image (right?), so we don't need this wrapper circle here

BrickheadJohnny (approved):

src/components/[guild]/EditGuild/components/Events/EventInput.tsx
Comment (zgen-marci) [reply src/components/[guild]/EditGuild/components/Events/EventInput.tsx:44]: If the link is null or undefined, we don't display the input. However, the link is an empty string which means the user wants to add a new link.

Comment (BrickheadJohnny) [reply src/components/[guild]/EditGuild/components/Events/EventInput.tsx:44]: ah okay, right, thanks!

src/components/[guild]/EditGuild/components/Events/EventInput.tsx
Comment (zgen-marci) [reply src/components/[guild]/EditGuild/components/Events/EventInput.tsx:50]: Yes, but the link3 logo looks misleading if there is no background.

image

PR #929, opened 2023-10-10
fix(edit guild): handle boolean fields properly

BrickheadJohnny:
We only checked if a form field value is truthy with a logical not operator inside `handleSubmitDirty`, so we actually filtered out boolean checkbox values & empty strings too (empty string is a valid form field value e.g. for a guild description), so I added a type check, let me know @cs-balazs if you think this is a good solution for the issue.


cs-balazs (approved):
Great fix! I agree that this is the proper way to handle this issue 🚀

PR #931, opened 2023-10-16
check dirty keys to keep when one of them is deleted.

zgen-marci:
The issue:
We can't delete the social and event links from the guild editor drawer.

Context and Expected Behaviour:
When we delete a link, we should send the rest to the backend because it will delete the links not in the request. 

The solution has two parts.
Part 1:
When we set the field to 'undefined' it won't be dirty. So I added an option to set them as dirty.

Part 2:
- We need to check the dirty fields to see if one of them is a key from 'DIRTY_KEYS_TO_KEEP'
- We set all the fields dirty to send to the backend.
- The rest of the current algorithm will filter the 'undefined' field as of now.
- As a result, we will send the fields not dirty as dirty with the current value and won't send the deleted.

Edge case:
The fix is not perfect, unfortunately, because it doesn't work when there is only one item in the list. It will send nothing.

### DIFF
--- a/src/utils/handleSubmitDirty.ts
+++ b/src/utils/handleSubmitDirty.ts
@@ -51,23 +51,8 @@ const formDataFilterForDirtyHelper = (dirtyFields: any, formData: any) => {
   }
 
   if (typeof dirtyFields === "object") {
-    const extendedDiryFields = JSON.parse(JSON.stringify(dirtyFields))
-
-    Object.entries(dirtyFields).forEach(([key]) => {
-      const keyIsDirtyFrom = DIRTY_KEYS_TO_KEEP.find(
-        (keeyToKeep) => keeyToKeep === key
-      )
-      const allfields = formData[keyIsDirtyFrom]
-
-      if (allfields)
-        Object.entries(allfields).forEach(([fieldKey, value]) => {
-          if (!extendedDiryFields[keyIsDirtyFrom].key)
-            extendedDiryFields[keyIsDirtyFrom][fieldKey] = value
-        })
-    })
-
     const newObj = Object.fromEntries(
-      Object.entries(extendedDiryFields)
+      Object.entries(dirtyFields)
         .map(([key, value]) => [
           key,
           formDataFilterForDirtyHelper(value, formData[key]),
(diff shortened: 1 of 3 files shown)


src/utils/handleSubmitDirty.ts
```diff
@@ -51,8 +51,23 @@ const formDataFilterForDirtyHelper = (dirtyFields: any, formData: any) => {
...
+
+    Object.entries(dirtyFields).forEach(([key]) => {
+      const keyIsDirtyFrom = DIRTY_KEYS_TO_KEEP.find(
+        (keeyToKeep) => keeyToKeep === key
+      )
+      const allfields = formData[keyIsDirtyFrom]
+
+      if (allfields)
+        Object.entries(allfields).forEach(([fieldKey, value]) => {
+          if (!extendedDiryFields[keyIsDirtyFrom].key)
```
cs-balazs, line 64: What is `.key` in this condition refer to?

zgen-marci: for example: 'eventSources' is the the 'keyIsDirtyFrom'  and 'eventBrite' is the .'key'

cs-balazs: Hmm, doesn't this solution cause things like `socialLinks` to be always sent?

zgen-marci: No, just in that case when some social links are dirty.

cs-balazs: Hmm I don't get why this condition is necessary, it is always true for form values that don't have a `key` field?

I'm thinking if the problem can be solved by modifying the `DIRTY_KEYS_TO_KEEP` logic a little:
```ts
DIRTY_KEYS_TO_KEEP.forEach((keyToKeep) => {
  if (dirtyFields[keyToKeep]) {
    newObj[keyToKeep] = formData[keyToKeep]
  }
})
```
and placing these lines right above the `return !isEmpty ? newObj : TO_FILTER_FLAG` seems to be working:
```ts
    const isEmpty = Object.keys(newObj).length <= 0

    // Re-attach id field to dirty objects
    if (!isEmpty && "id" in formData) {
      newObj.id = formData.id
    }
```
It also seems to be working fine for the case when I remove the last social link

cs-balazs (approved):

PR #936, opened 2023-10-19
Queues with feature flag

Devidxyz:


src/utils/createAndAwaitJob.ts
```diff
@@ -0,0 +1,41 @@
...
+) {
+  const poll = (): Promise<Job[]> =>
+    fetcherWithSign([
+      `${url}?${new URLSearchParams(queryParams).toString()}`,
+      { method: "GET" },
+    ])
+
+  const initialPollResult = await poll().catch(() => null as Job[])
+
+  const jobAlreadyInProgress = initialPollResult?.find((job) => !job.done)
```
BrickheadJohnny, line 17: Will this work properly? I mean what if a job is in progress, but the user starts another job (e.g. another join flow)? Shouldn't we distinguish jobs somehow? 👀 

(btw it's good for now I think, since we'll only enable this feature in one guild)

BrickheadJohnny (commented):
Looks good to me overall, left only 1 comment in `createAndAwaitJob` ;)

I think we need the `(... as any)` casts because we're missing some types in the `@guildxyz/types` package, right? Let's don't forget to remove these casts once those types are available in the package.

Also, quick note: don't forget to revert the `.env` changes & remove `: any` from the feature flag before merging the PR!

src/utils/createAndAwaitJob.ts
BrickheadJohnny: Yes, I think it'll be fine this way then, thanks!

PR #946, opened 2023-11-02
feat: capacity & time for rewards

BrickheadJohnny:


src/platforms/ContractCall/ContractCallReward.tsx
```diff
@@ -57,7 +62,9 @@ const ContractCallReward = ({
           </LinkButton>
         </>
       }
-    />
+    >
+      {showCapacityTimeTags && <CapacityTimeTags rolePlatform={platform} />}
```
dovalid, line 66: Same as above

src/components/[guild]/RolePlatforms/components/EditRolePlatformCapacityTimeModal.tsx
```diff
@@ -0,0 +1,251 @@
...
+  if (!datetimeLocal) return undefined
+
+  try {
+    return new Date(datetimeLocal).toISOString()
+  } catch {
+    return undefined
+  }
+}
+
+const EditRolePlatformCapacityTimeModal = ({
```
Comment (dovalid) [inline src/components/[guild]/RolePlatforms/components/EditRolePlatformCapacityTimeModal.tsx:60]: Would be nice to use Availibilty instead of CapacityTime. Not just here but everywhere, except CAPACITY_TIME_PLATFORMS maybe and when we specifically refer to either capacity or time ofc. Availibilty is a nice collective term for them imo

src/platforms/ContractCall/ContractCallCardMenu.tsx
```diff
@@ -1,14 +1,65 @@
...
+      <PlatformCardMenu>
+        {rolePlatform && (
+          <MenuItem icon={<Clock />} onClick={onOpen}>
+            Edit availibility
+          </MenuItem>
+        )}
+        <RemovePlatformMenuItem platformGuildId={platformGuildId} />
+      </PlatformCardMenu>
+
+      <EditRolePlatformCapacityTimeModal
```
dovalid, line 49: It would be nice to make an EditRewardAvailibilityMenuItem component similar to RemovePlatformMenuItem and just use that in ContractCardMenu, UniqueTextCardMenu and SecretTextCardMenu

src/components/[guild]/RolePlatforms/RolePlatforms.tsx
```diff
@@ -124,6 +139,45 @@ const RolePlatforms = ({ roleId }: Props) => {
...
                   }
                   actionRow={PlatformCardSettings && <PlatformCardSettings />}
+                  contentRow={
+                    CAPACITY_TIME_PLATFORMS.includes(type) && (
+                      <Wrap>
+                        {showCapacityTimeTags && (
+                          <CapacityTimeTags rolePlatform={rolePlatform} />
+                        )}
+
+                        <EditRolePlatformCapacityTimeButton
```
Comment (dovalid) [inline src/components/[guild]/RolePlatforms/RolePlatforms.tsx:149]: There're invidual buttons and modals mounted for each reward now, but only a single state, so if there're multiple CAPACITY_TIME supporting rewards in a role, and I click on edit, every modal is going to open. I think we should abstract the modal and the state into EditRewardAvailibilityButton (EditRolePlatformCapacityTimeButton now), or as I'm seeing there's CapacityTimeSetup that abstracts just something like that, why not use that?

src/components/[guild]/AccessHub/hooks/useEditGuildPlatform.ts
```diff
@@ -32,6 +33,30 @@ const useEditGuildPlatform = ({
             if (gp.id === guildPlatformId) return response
             return gp
           }),
+          roles: CAPACITY_TIME_PLATFORMS.includes(
+            PlatformType[response.platformId] as PlatformName
+          )
+            ? prevGuild.roles.map((role) => {
```
Comment (dovalid) [inline src/components/[guild]/AccessHub/hooks/useEditGuildPlatform.ts:39]: Altough it's only used in two places, I think we should abstract it into a util function (not necessarily in the utils folder, just in one of the files and export it from there), it's complex enough and would be more self-explaining too if we label it somehow

dovalid (changes requested):

src/components/[guild]/RolePlatforms/RolePlatforms.tsx
```diff
@@ -124,6 +139,45 @@ const RolePlatforms = ({ roleId }: Props) => {
                     )
                   }
                   actionRow={PlatformCardSettings && <PlatformCardSettings />}
+                  contentRow={
+                    CAPACITY_TIME_PLATFORMS.includes(type) && (
+                      <Wrap>
```
Comment (dovalid) [inline src/components/[guild]/RolePlatforms/RolePlatforms.tsx:144]: Here's an external Wrap, and there's another wrap in CapacityTimeTags. Would be nice if both the tags and the edit button could be in a single Wrap

src/platforms/components/TextReward.tsx
```diff
@@ -100,7 +129,9 @@ const SecretTextReward = ({ platform, withMotionImg }: RewardProps) => {
             </Tooltip>
           )
         }
-      />
+      >
+        {showCapacityTimeTags && <CapacityTimeTags rolePlatform={platform} />}
```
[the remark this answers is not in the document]
BrickheadJohnny: Good point, removed!

src/platforms/UniqueText/UniqueTextCardMenu.tsx
```diff
@@ -10,19 +14,46 @@ type Props = {
...
+    },
+  })
+
   return (
     <>
       <PlatformCardMenu>
         <MenuItem icon={<PencilSimple />} onClick={onOpen}>
           Edit unique secret
         </MenuItem>
+        {rolePlatform && (
```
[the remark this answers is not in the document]
BrickheadJohnny: I thought it might be undefined if someone just adds a guildPlatform with our API endpoint, but actually we should handle this (= disable it) on our backend, so I removed this condition in https://github.com/agoraxyz/guild.xyz/pull/946/commits/d8ae2dabdee818881ed30dfad3c8444c40c8efdb & https://github.com/agoraxyz/guild.xyz/pull/946/commits/5b4132c784ecffc31d27a6b9b69980202aec6e22!

src/platforms/ContractCall/ContractCallCardMenu.tsx
BrickheadJohnny: Nice, I completely forgot about this copypasta, fixed in https://github.com/agoraxyz/guild.xyz/pull/946/commits/5b4132c784ecffc31d27a6b9b69980202aec6e22!

src/components/[guild]/RolePlatforms/RolePlatforms.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/RolePlatforms/RolePlatforms.tsx]: Uh, awesome idea, thanks, made a refactor in https://github.com/agoraxyz/guild.xyz/pull/946/commits/927de57a884fa0f7fe50e12bc7ddfc5633f58863! 👌 I think it became much cleaner, and I also managed to remove the outer `Wrap`s in a couple of places, let me know if you like it this way!

dovalid (approved):

PR #949, opened 2023-11-05
Analytics

dovalid:


src/components/analytics/MembersChart/MembersChart.tsx
```diff
@@ -0,0 +1,205 @@
...
+    setShownLines(["total", ...roles?.map((role) => role.id.toString())])
+  }, [roles])
+
+  const roleColors: Record<number, string> = useMemo(
+    () =>
+      roles?.reduce((acc, curr) => {
+        acc[curr.id] =
+          /* curr.imageUrl
+          ? await getColorByImage(curr.imageUrl)
+          : */ Object.values(allColors)[getRandomInt(9)][4]
```
BrickheadJohnny, line 80: I think this should be `getRandomInt(12)`, right? There are 13 different colors inside `allColors`, and each color array has 9 hex codes inside it (I guess that's why you selected the 4th index) 👀

BrickheadJohnny (approved):
Left only one comment and also added some types / fixed the strange type error 😉 🚀

PR #950, opened 2023-11-07
feat: fuel wallet support & fuel requirements

BrickheadJohnny:
- added Fuel & Fuelet wallet support
- added some new state in `useWeb3ConnectionManager`: isWeb3Connected (EVM or Fuel wallet is connected), address (EVM or Fuel address), type (`EVM` or `FUEL`), disconnect (for handling wallet disconnection, regardless of which type of wallet is connected)
- added a `useFuel` hook for managing the Fuel connection state & lazy loading the Fuel SDK
- added a new requirement type for Fuel (contains token balance for now, but we'll add more requirement types later, that's why I added a select input for it)

Relevant PRs:
- Core: https://github.com/agoraxyz/guild-backend/pull/1113
- Gate: https://github.com/agoraxyz/guild-gate/pull/635

Deployment checklist:
- [x] put app in maintenance mode
- [x] switch to prod api & use the prod captcha key
- [x] check if the test are passing with the prod API
- [x] migrate the db & merge the core/gate PRs
- [x] merge this PR


src/components/_app/FuelSetup.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { atom, useAtom } from "jotai"
+import { useEffect } from "react"
+
+export const fuelConnectedAtom = atom(false)
+export const fuelConnectingAtom = atom(false)
+export const fuelAddressAtom = atom("" as `0x${string}`)
+export const fuelWalletAtom = atom(null as WalletUnlocked)
+export const fuelProviderAtom = atom(null as Provider)
+
+const FuelSetup = () => {
```
cs-balazs, line 14: This might be a silly question, but is there a reason this is a component and not a hook? 😄

src/components/_app/KeyPairProvider.tsx
```diff
@@ -265,6 +266,7 @@ const KeyPairProvider = ({ children }: PropsWithChildren<unknown>): JSX.Element
               "You've connected your account from a new device, so you have to sign a new message to stay logged in",
             duration: 5000,
           })
+          openWalletSelectorModal()
```
cs-balazs, line 269: You showed this IRL today, is it still needed, knowing there are cases when this doesn't run, but the modal is needed (when there is no previous keypair, which has been invalidated)

src/hooks/useSubmit/useSubmit.ts
```diff
@@ -219,18 +236,70 @@ const useSubmitWithSign = <ResponseType>(
...
+  const params = createMessageParams(address, ts, msg, payload)
+  let sig = null
+
+  if (!!keyPair && !forcePrompt) {
+    params.method = ValidationMethod.KEYPAIR
+    sig = await signWithKeyPair(keyPair, params)
+  } else {
+    params.method = ValidationMethod.STANDARD
+    sig = await wallet.signMessage(getMessage(params))
+  }
```
cs-balazs, line 298: Hmmm, on the BE there is a new auth method, but here the same one seems to be used as for an EVM wallet :thinking:
The new one doesn't even seem to be defined on the enum, maybe the new method is completely unnecessary, and we only need the validator changes to accept a Fuel address on this method?

cc @marclllaks

src/components/common/Layout/components/Account/components/AccountModal/components/LinkAddressButton.tsx
```diff
@@ -40,6 +40,7 @@ const LinkAddressButton = (props) => {
 
     try {
       await walletClient.requestPermissions({ eth_accounts: {} })
+    } catch {
```
cs-balazs, line 43: Just out of curiosity, does this change have an effect, or just left here by accident? 😄 (I'm actually not really sure, if the `catch` is missing, the error gets caught or not :harold:)

cs-balazs (commented):
Awesome PR 🚀 
I'll approve as soon as I can try the preview, I might need a bit of help with that 👉 👈

src/components/[guild]/Requirements/components/GuildCheckout/components/buttons/ConnectWalletButton.tsx
```diff
@@ -1,23 +1,28 @@
...
-        colorScheme="blue"
-        onClick={openWalletSelectorModal}
-        isDisabled={isConnected}
-        w="full"
+    <Collapse in={type !== "EVM"}>
+      <Tooltip
+        label="Disconnect your Fuel wallet first"
+        isDisabled={!isWeb3Connected}
+        hasArrow
       >
```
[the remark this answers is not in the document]
Comment (BrickheadJohnny) [reply src/components/[guild]/Requirements/components/GuildCheckout/components/buttons/ConnectWalletButton.tsx:15]: Yup, it should be visible, because currently we do not support any on-chain actions on Fuel, only on EVM chains, so this acts as a "Connect EVM wallet" button (maybe we should update the copy? 👀)

src/components/_app/FuelSetup.tsx
BrickheadJohnny: This whole logic should only run once, and it ran around 60 times when it was in the `useFuel` hook 😂 Actually, that was a problem only when we used the `checkConnection` function, since it was called 60+ times.

I think another solution would be to put these "heavy" functions (or just functions which should run only once) in an SWR call, I don't know if that'd be better or not, what do you think?

src/components/common/Layout/components/Account/components/AccountModal/components/LinkAddressButton.tsx
BrickheadJohnny: I added the catch because the error didn't get caught haha (and it's a completely valid case that the user just clicks on "Cancel" inside the MetaMask popup - well, that caused a pretty ugly error :harold:)

src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/ConnectorButton.tsx
```diff
@@ -79,8 +74,6 @@ const ConnectorButton = ({
       w="full"
       size="xl"
       justifyContent="space-between"
-      border={activeConnector?.id === connector.id && "2px"}
-      borderColor="primary.500"
     >
       {!isConnected || !(activeConnector?.id === connector.id)
         ? `${connectorName}`
```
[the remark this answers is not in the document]
BrickheadJohnny: Tru dat, fixed in: https://github.com/agoraxyz/guild.xyz/pull/950/commits/3c6761b2bce64f0d5607ee49362034d389f8d05a

src/components/_app/KeyPairProvider.tsx
BrickheadJohnny: Nice catch, we don't need this line anymore! Fixed in: https://github.com/agoraxyz/guild.xyz/pull/950/commits/0e2d3b4aa9a7b3fc8b5da47c0c0a78aad058d424

src/components/_app/FuelSetup.tsx
BrickheadJohnny: I ended up moving this logic back to the `useFuel` hook, since we don't use auto connection, so now all the related logic is in the same file: https://github.com/agoraxyz/guild.xyz/pull/950/commits/c87e9d7b86a04506c03f1ec7e1282c3e2e08def5

src/components/[guild]/Requirements/components/GuildCheckout/components/buttons/ConnectWalletButton.tsx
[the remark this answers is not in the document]
Comment (cs-balazs) [reply src/components/[guild]/Requirements/components/GuildCheckout/components/buttons/ConnectWalletButton.tsx:15]: Aah, I see, I don't think we need to change it, I was just making sure I understand the change 👍

src/components/_app/FuelSetup.tsx
cs-balazs: Nice! Thanks for the change 🚀

src/components/common/Layout/components/Account/components/AccountModal/components/LinkAddressButton.tsx
cs-balazs: Wow, nice to know it works like that, would have guessed, that it still catches, just doesn't do anything with it 🤔 I guess it makes sense since a `somePromise.then(...).finally(...)` would also reject
Thanks, I learned something today 😄

src/hooks/useSubmit/useSubmit.ts
BrickheadJohnny: Very nice catch, we'll use the `FUEL` auth method from now: https://github.com/agoraxyz/guild.xyz/pull/950/commits/a8e03d8ff5009db3b52058984fd3a99d3385f568

PR #956, opened 2023-11-10
Web3ConnectionManager refactor with Jotai

BrickheadJohnny:
I noticed that we actually use 2 values from this context inside KeyPairProvider, which is above Web3ConnectionManager, so those values were always undefined. Since we only store simple booleans and strings (and these values should be global) in this context, I thought we could refactor it and just use Jotai for state management.

This refactor will also come handy in the #950 PR


src/components/_app/Web3ConnectionManager/hooks/useWeb3ConnectionManager.ts
```diff
@@ -0,0 +1,141 @@
...
+import { PlatformName } from "types"
+import { useAccount, useSwitchNetwork } from "wagmi"
+
+const delegateConnectionAtom = atom(false)
+const safeContextAtom = atom(false)
+const walletSelectorModalAtom = atom(false)
+const accountModalAtom = atom(false)
+const platformMergeAlertAtom = atom(false)
+const accountMergeAddressAtom = atom("")
+const accountMergePlatformNameAtom = atom("" as PlatformName)
```
cs-balazs, line 15: Would it make sense to export these atoms instead of returning them from the hook (I mean the ones where only the value and the setter are returned)

cs-balazs (approved):
Nice 🚀 
Left one minor question :ayy:

src/components/_app/Web3ConnectionManager/hooks/useWeb3ConnectionManager.ts
BrickheadJohnny: Note: we talked about it irl, we'll leave it as is for now

dovalid (approved):

PR #960, opened 2023-11-15
feat: scroll supports

daominhhiep:
- fix explorer scroll alpha
- added support for these scroll sepolia and scroll mainnet


BrickheadJohnny:
Hey @daominhhiep, thank you for your PR! I'll merge it to another dev branch now, where we'll add a couple of other chains too, then we'll ship all new chains at once 🚀

PR #966, opened 2023-11-20
feat(requirements): custom image & text

zgen-marci:


src/components/[guild]/Requirements/components/Requirement.tsx
```diff
@@ -35,76 +43,99 @@ const Requirement = ({
...
+                  <RequirementButton
+                    rightIcon={<Icon as={CaretDown} />}
+                    onClick={togglePreview}
+                  >
+                    View original
+                  </RequirementButton>
+                </PopoverTrigger>
+                <Portal>
+                  <PopoverContent minW="max-content">
+                    <OriginalRequirementPreview
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/components/Requirement.tsx:120]: I think we could completely get rid of this component and use the `RequirementDisplayComponent` instead with `footer={null}` and `rightElement={!!fieldRoot ? <ResetRequirementButton /> : null}`

src/components/[guild]/Requirements/components/RequirementImage.tsx
```diff
@@ -0,0 +1,52 @@
+import { Circle, Img, SkeletonCircle, useColorMode } from "@chakra-ui/react"
+
+type Props = {
+  isImageLoading?: boolean
+  image?: string | JSX.Element
+  withImgBg?: boolean
+}
+
+const RequirementImage = ({ isImageLoading, withImgBg, image }: Props) => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/components/RequirementImage.tsx:9]: I think if we remove the `OriginalRequirementPreview` component as I mentioned in my other comment, we won't need this one, right? We could just conditionally render the image editor inside `Requirement.tsx`. What do you think?

src/components/[guild]/Requirements/components/RequirementImageEditor.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <FormControl isInvalid={!!fileRejections?.[0]}>
+      <Wrap>
+        {customImage ? (
+          <Box position={"relative"}>
+            <Circle
+              position="absolute"
+              onClick={() => {
+                setValue(`requirements.${index}.data.customImage`, undefined, {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/components/RequirementImageEditor.tsx:66]: We might need to set this to empty string, but not sure about it, could you please test it?

src/components/[guild]/Requirements/components/RequirementImage.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/Requirements/components/RequirementImage.tsx:9]: Nevermind, now I see we can use this component in 3 different places!

PR #969, opened 2023-11-27
fix(useSWRWithOptionalAuth): use mutate properly

BrickheadJohnny:
The data wasn't invalidated properly in `useYourGuilds` after successful guild creation, so I modified the mutator function inside `useSWRWithOptionalAuth`


cs-balazs (approved):
Nice change 🚀

PR #972, opened 2023-11-29
[GUILD-1884] - Use separate POST endpoint for address linking

cs-balazs:
[[LINEAR]](https://linear.app/guildxyz/issue/GUILD-1884/use-the-post-addresses-for-address-linking-on-the-fe)

This PR aims to simplify the address linking flow by using the dedicated `POST /v2/users/:id/addresses` endpoint. The whole logic is now contained within the `useLinkAddress` hook which is the `useSubmit` + one `useEffect`, which is explained in the comment above it. The state is still managed by the Jotai atom as before, but its usage got (imo) simpler


src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/hooks/useLinkAddress.ts
```diff
@@ -0,0 +1,111 @@
...
+  address: `0x${string}`,
+  signMessageAsync: (
+    args?: Parameters<ReturnType<typeof useSignMessage>["signMessageAsync"]>[0]
+  ) => Promise<`0x${string}`>
+) => {
+  const addr = address.toLowerCase()
+  const timestamp = Date.now()
+  const nonce = randomBytes(32).toString("hex")
+  const message = `Address: ${addr}\nNonce: ${nonce}\n Timestamp: ${timestamp}`
+  const signature = await signMessageAsync({ message })
```
BrickheadJohnny, line 28: I think this won't work with Fuel :( But we could use `wallet` from the `useFuel` hook, and then sign message this way
```ts
const { type } = useWeb3ConnectionManager()
const { wallet: fuelWallet } = useFuel()
const signature = type === "FUEL" ? await wallet.signMessage() : await signMessageAsync({ message })
```

Or lmk if you have a cleaner idea for this, because it'll get really messy if we add more and more wallet types in the future :ayy:

(My other idea is to just provide a `signMessage` function from the `useWeb3ConnectionManager` hook, so we could handle the wallet type related logic there)

cs-balazs: Nice catch! Completely forgot about Fuel :harold:
I like the idea of a `signMessage` function inside `useWeb3ConnectionManager`, made this change in https://github.com/agoraxyz/guild.xyz/pull/972/commits/a185961fd70ab5e3e7c6c7c5710177d1d00da797

cs-balazs: The PR now depends on https://github.com/agoraxyz/guild-backend/pull/1162, as the backend didn't support Fuel either on this endpoint

BrickheadJohnny: Great, thank you, approved! 🚀

PR #973, opened 2023-11-30
Fix/guild creation flow

zgen-marci:
A fix for

[Captcha missing data](https://linear.app/guildxyz/issue/GUILD-1926/captcha)
[Limit supply](https://linear.app/guildxyz/issue/GUILD-1927/limit-supply-nft)

### DIFF
--- a/src/components/create-guild/templates.ts
+++ b/src/components/create-guild/templates.ts
@@ -23,9 +23,7 @@ export const TEMPLATES: Array<RoleFormType> = [
     requirements: [
       {
         type: "CAPTCHA",
-        data: {
-          maxAmount: 864000000,
-        },
+        data: {},
       },
     ],
   },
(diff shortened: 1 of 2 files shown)


src/components/create-guild/templates.ts
```diff
@@ -23,6 +23,9 @@ export const TEMPLATES: Array<RoleFormType> = [
     requirements: [
       {
         type: "CAPTCHA",
+        data: {
+          maxAmount: 864000000,
```
BrickheadJohnny, line 27: I think this can be an empty object (so the requirement will be to complete a captcha, without a time limit) 👍

BrickheadJohnny (approved):
Left 1 comment, but otherwise lgtm! 🚀

src/components/create-guild/templates.ts
zgen-marci: In that case, It seems the user is verified instantly without a captcha. It is an expected behavior? 

image

BrickheadJohnny: If you've already completed any captcha on guild, it is expected, yes ;)

PR #977, opened 2023-12-06
Don't filter SendReward/RevokeReward actions in audit log

Devidxyz:
Ready to merge once we use v2 joins in every* guild


BrickheadJohnny (approved):
wow what a beautiful pr

Devidxyz:
I think the time has come, this can be merged, what do you think @BrickheadJohnny, it won't break anything, right?

BrickheadJohnny:
> I think the time has come, this can be merged, what do you think @BrickheadJohnny, it won't break anything, right?

LFG @Devidxyz

PR #978, opened 2023-12-06
feat: Visit link requirement

BrickheadJohnny:


src/components/[guild]/Requirements/components/Requirement.tsx
```diff
@@ -69,7 +63,9 @@ const Requirement = ({
       <VStack alignItems={"flex-start"} alignSelf="center" spacing={1.5}>
         <ChildrenWrapper {...wrapperProps} display="inline-block">
           {requirement?.isNegated && <Tag mr="2">DON'T</Tag>}
-          {requirement?.data?.customName || children}
+          {requirement?.type === "LINK_VISIT"
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/components/Requirement.tsx:66]: I don't really like this, had another (more generalized) idea, but I ended up using this approach. More details in my PR comment in `VisitLinkRequirement.tsx`

src/requirements/VisitLink/VisitLinkRequirement.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <Link {...linkProps}>{data.id}</Link>
+    </>
+  )
+
+  return (
+    <Requirement
+      image={<Icon as={LinkIcon} boxSize={6} />}
+      {...props}
+      showViewOriginal={false}
+      footer={
```
BrickheadJohnny, line 75: So I basically overwrote the default "View original" popover here, because this requirement type won't ever have a real default value (see `Requirement.tsx`), but let me explain the other solution which I had in my mind:

We could add an optional `childrenWrapperComponent` prop to `requirements.ts` and make a `VisitLinkChildrenParser` component which would to the following stuff:
- check if there's a string child
- if there isn't, just return children
- if there is, it could try to parse it (see lines 45-47 & 97-105 in this file) and display the parsed value

Not sure which would be the cleaner solution, that's why I sticked with the first one... but lmk what's your opinion on this!

PR #981, opened 2023-12-11
feat: PolygonID minting

zgen-marci:


src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddPolygonIdPanel.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  const { append } = useFieldArray({
+    name: "rolePlatforms",
+  })
+
+  useEffect(() => {
+    append({
+      guildPlatform: {
+        platformName: "POLYGON_ID",
+        platformGuildId: PlatformType.POLYGON_ID.toString(),
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddPolygonIdPanel.tsx:20]: this won't be unique for each guild, let's use `${PlatformType.POLYGON_ID}-${guildId}` instead

src/platforms/PolygonId/components/NoDID.tsx
```diff
@@ -0,0 +1,89 @@
...
+    formState: { errors },
+    control,
+    handleSubmit,
+  } = useForm({ defaultValues: { did: "" } })
+
+  const DID = useWatch({ name: "did", control })
+
+  return (
+    <form
+      onSubmit={handleSubmit(({ did }) =>
```
BrickheadJohnny, line 42: We can remove this `<form>` wrapper, and just add the `onSubmit` handler on the button below

src/platforms/PolygonId/components/NoDID.tsx
```diff
@@ -0,0 +1,89 @@
...
+import FormErrorMessage from "components/common/FormErrorMessage"
+import { useSubmitWithSign } from "hooks/useSubmit"
+import { Stack } from "phosphor-react"
+import { useForm, useWatch } from "react-hook-form"
+import { mutate } from "swr"
+import fetcher from "utils/fetcher"
+
+const NoDID = () => {
+  const { id } = useUser()
+  const connectPolygonId = useSubmitWithSign(
```
BrickheadJohnny, line 23: For a bit cleaner code, we could destructure this here, and just use `onSubmit` and `isLoading` from it

src/platforms/PolygonId/components/NoDID.tsx
```diff
@@ -0,0 +1,89 @@
...
+          <FormErrorMessage>{errors?.did?.message}</FormErrorMessage>
+        </FormControl>
+        <Button
+          isDisabled={DID === ""}
+          colorScheme="green"
+          mt={8}
+          ml="auto"
+          type="submit"
+          isLoading={connectPolygonId.isLoading}
+          loadingText={"connecting..."}
```
BrickheadJohnny, line 80: Capital initial letter ;)

src/platforms/PolygonId/components/PolygonIdQRCode.tsx
```diff
@@ -0,0 +1,139 @@
...
+            </Text>
+          </Box>
+        </HStack>
+      </ModalHeader>
+      <ModalBody pt={8}>
+        <Center flexDirection={"column"}>
+          {claim.error || qr.error ? (
+            <ErrorAlert
+              label={
+                qr.error ? "Couldn't generate QR code" : "Couldn't mint the proof"
```
BrickheadJohnny, line 99: I think `Couldn't mint PolygonID proof` would look better here

src/platforms/PolygonId/components/PolygonIdQRCode.tsx
```diff
@@ -0,0 +1,139 @@
...
+                leftIcon={<ArrowsClockwise />}
+                isLoading={qr.isLoading}
+                loadingText={"Generating QR code"}
+                color="gray"
+                onClick={() => mutate(QR_URL)}
+              >
+                Generate new QR code
+              </Button>
+              <Text mt="10" textAlign="center">
+                Scan with your Polygon ID app! The modal will automatically close on
```
BrickheadJohnny, line 128: This is not true anymore, so we can remove it

src/platforms/PolygonId/usePolygonIdCardProps.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  return {
+    name: "PolygonID",
+    type: "POLYGON_ID" as PlatformName,
+    image: platformGuildData?.imageUrl ?? (
+      <Circle size={10} bgColor={bgColor}>
+        <Img src={platforms.POLYGON_ID.imageUrl} boxSize={10} color="white" />
+      </Circle>
+    ),
+    info: rolePlatform && <AvailibiltyTags rolePlatform={rolePlatform} mt={1} />,
```
BrickheadJohnny, line 25: PolygonID won't have capacity/time, so we don't need to return `info` here

src/platforms/components/PolygonIDPreview.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const PolygonIDPreview = (): JSX.Element => {
+  const platformGuildData = useWatch({
+    name: "rolePlatforms.0.guildPlatform.platformGuildData",
+  })
+
+  return (
+    <PlatformPreview
+      type="POLYGON_ID"
+      name={platformGuildData?.name ?? "PolygonID"}
```
BrickheadJohnny, line 13: We won't update `platformGuildData.name`, so let's just use `"PolygonID"` here

src/platforms/components/PolygonIDPreview.tsx
```diff
@@ -0,0 +1,19 @@
...
+      image={platformGuildData?.imageUrl ?? platforms.POLYGON_ID.imageUrl}
```
BrickheadJohnny, line 14: Same here, just use `platforms.POLYGON_ID.imageUrl` please

src/platforms/components/PolygonIdReward.tsx
```diff
@@ -0,0 +1,113 @@
...
+import useIsMember from "components/[guild]/hooks/useIsMember"
+import Button from "components/common/Button"
+import { ArrowSquareOut, LockSimple } from "phosphor-react"
+import { MintPolygonIdProofModal } from "platforms/PolygonId/components/MintPolygonIdProofModal"
+import platforms from "platforms/platforms"
+import { useMemo } from "react"
+import { PlatformType } from "types"
+import { useAccount } from "wagmi"
+
+const PolygonIdReward = ({ platform, withMotionImg }: RewardProps) => {
```
BrickheadJohnny, line 22: We can remove capacity/time related stuff here, that's only applicable for text-based rewards & NFT rewards

src/platforms/PolygonId/components/PolygonIdMintingModalContent.tsx
```diff
@@ -0,0 +1,105 @@
...
+    <Stack>
+      <AlertDescription position="relative" top={0.5} fontWeight="semibold" pr="4">
+        Server not available
+      </AlertDescription>
+    </Stack>
+    <Spacer />
+  </Alert>
+)
+
+const PolygonIdMintingModalContent = () => {
```
BrickheadJohnny, line 46: This is not a too complex component, and I think we don't use it in multiple files, so in my opinion we can move this logic&markup into `MintPolygonIdProofModal`

BrickheadJohnny (changes requested):

src/platforms/PolygonId/components/NoDID.tsx
```diff
@@ -0,0 +1,88 @@
...
+  Input,
+  ModalBody,
+  ModalHeader,
+  Spacer,
+} from "@chakra-ui/react"
+import useUser from "components/[guild]/hooks/useUser"
+import Button from "components/common/Button"
+import FormErrorMessage from "components/common/FormErrorMessage"
+import { useSubmitWithSign } from "hooks/useSubmit"
+import { Stack } from "phosphor-react"
```
BrickheadJohnny, line 16: I'm pretty sure this should be imported from Chakra UI ;)

src/platforms/PolygonId/components/NoDID.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useForm, useWatch } from "react-hook-form"
+import { mutate } from "swr"
+import fetcher from "utils/fetcher"
+
+const NoDID = () => {
+  const { id } = useUser()
+  const { isLoading, onSubmit, error } = useSubmitWithSign(
+    () =>
+      fetcher(`${process.env.NEXT_PUBLIC_POLYGONID_API}/v1/polygon-id/connect`, {
```
BrickheadJohnny, line 25: This probably won't work this way, you should do something like this (pass down `signedValidation` too:
```ts
(signedValidation: SignedValdation) =>
  fetcher(`${process.env.NEXT_PUBLIC_POLYGONID_API}/v1/polygon-id/connect`, {
    ...signedValidation,
    method: "POST",
  })
```

src/platforms/PolygonId/components/NoDID.tsx
```diff
@@ -0,0 +1,88 @@
...
+        method: "POST",
+      }),
+    {
+      onSuccess: () => {
+        mutate(`${process.env.NEXT_PUBLIC_POLYGONID_API}/v1/users/${id}/polygon-id`)
```
BrickheadJohnny, line 30: This could be replaced with an optimistic update as I mentioned in my other comment in `MintPolygonIdProofModalContent.tsx`

src/platforms/PolygonId/components/NoDID.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <FormLabel>Paste your DID</FormLabel>
+          <Input
+            {...register("did", {
+              required: "This field is required",
+            })}
+          />
+          <FormErrorMessage>{errors?.did?.message}</FormErrorMessage>
+        </FormControl>
+        <Button
+          isDisabled={DID === ""}
```
BrickheadJohnny, line 73: Couldn't we use the shorted `!DID` here? 👀

src/platforms/PolygonId/components/PolygonIdQRCode.tsx
```diff
@@ -0,0 +1,114 @@
...
+  goBack: () => void
+}
+
+const PolygonIdQRCode = ({ role, goBack }: Props) => {
+  const { id: userId } = useUser()
+  const { id: guildId } = useGuild()
+  const QR_URL = `${process.env.NEXT_PUBLIC_POLYGONID_API}/v1/users/${userId}/polygon-id/claim/${guildId}:${role.id}/qrcode`
+
+  const claim = useSubmitWithSign(() =>
+    fetcher(`${process.env.NEXT_PUBLIC_POLYGONID_API}/v1/polygon-id/claim`, {
```
BrickheadJohnny, line 34: This will probably also need a `signedValidation`

src/platforms/PolygonId/usePolygonIdCardProps.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const usePolygonIdCardProps = (guildPlatform: GuildPlatform) => {
+  const bgColor = useColorModeValue("gray.700", "gray.600")
+
+  const platformGuildData = guildPlatform.platformGuildData
+
+  return {
+    name: "PolygonID proofs",
+    type: "POLYGON_ID" as PlatformName,
+    image: platformGuildData?.imageUrl ?? (
```
BrickheadJohnny, line 13: Users won't be able to modify the platform's image, so you can just use `platforms.POLYGON_ID.imageUrl` here

BrickheadJohnny (changes requested):

src/platforms/PolygonId/components/PolygonIdQRCode.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const claim = useSubmitWithSign(submit, {
+    onSuccess: () => setShouldFetchQR(QR_URL),
+  })
+  const qr = useSWR(shouldFetchQR)
+
+  const qrSize = useBreakpointValue({ base: 300, md: 400 })
+
+  useEffect(() => {
+    if (userId && guildId && !isClaimed) {
+      claim.onSubmit({
```
BrickheadJohnny, line 51: This onSubmit should happen conditionally when the user presses the "Mint proof" button on a `MintableRole` card. So in short, I think we should check if there's a QR code for a role and if we can fetch it, we just display that QR in a modal when the user clicks on the "Show QR code" button. If there's no QR code, the button's label is "Mint proof" and if the user clicks on it, we fire this onSubmit & also open the modal in a loading state.

src/platforms/PolygonId/components/MintPolygonIdProofModalContent.tsx
```diff
@@ -0,0 +1,112 @@
...
+      ? `${process.env.NEXT_PUBLIC_POLYGONID_API}/v1/users/${userId}/polygon-id`
+      : null,
+    null,
+    {
+      onErrorRetry: (error) => {
+        if (error.status === 500) return
+      },
+    }
+  )
+  const claimedRoles = useSWR(
```
BrickheadJohnny, line 62: We could use `useSWRImmutable` here and I'd also move the whole `isClaimed` logic into the `MintableRole` component instead of passing down props.

src/platforms/PolygonId/components/MintableRole.tsx
```diff
@@ -0,0 +1,46 @@
+import { Card, HStack, Heading, Spacer } from "@chakra-ui/react"
+import Button from "components/common/Button"
+import GuildLogo from "components/common/GuildLogo"
+import { Role } from "types"
+
+type Props = {
```
BrickheadJohnny, line 6: We could simplify this component to have only 1 prop (`role`), see my other comments for more context ;)

src/platforms/PolygonId/components/MintPolygonIdProofModalContent.tsx
```diff
@@ -0,0 +1,112 @@
...
+      (rolePlatform) => rolePlatform.guildPlatformId === guildPlatformId.id
+    )
+  }
+
+  const isClaimed = (role: Role) =>
+    claimedRoles.data.length &&
+    !!claimedRoles.data[0].roleIds.find((claimedRoleId) => claimedRoleId === role.id)
+
+  return mint ? (
+    <PolygonIdQRCode
```
BrickheadJohnny, line 87: I think we could move this modal into `MintableRole` and open it from there, no need to conditionally switch the content of this modal.

src/platforms/PolygonId/components/MintPolygonIdProofModalContent.tsx
```diff
@@ -0,0 +1,112 @@
...
+    `${process.env.NEXT_PUBLIC_POLYGONID_API}/v1/users/${userId}/polygon-id/claims?format=role&guildId=${guildId}`
+  )
+
+  const [mint, setMint] = useState<Role | null>(null)
+
+  if (checkDID.isLoading || claimedRoles.isLoading) return <LoadingState />
+
+  if (checkDID.error) return <NoDID />
+
+  const onlyWithPolygonIdReward = (role: Role) => {
```
BrickheadJohnny, line 72: This way the `guildPlatforms.find...` line will run for each role, I think we can refactor it this way:
- move the `const guildPlatformId = ...` line outside of the function
- do the filter logic inside the `roles.filter(...)` line, since it's not a complicated logic and we only use it there

BrickheadJohnny (changes requested):
Left a couple of comments! Please also double-check if there are any other places where you could use `useSWRImmutable` instead of `useSWR` to avoid unnecessary API calls.

PR #983, opened 2023-12-11
feat(CRM): Customize columns

dovalid:


src/components/[guild]/crm/CustomizeViewModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+}
+
+const CustomizeViewModal = ({ isOpen, onClose, table }: Props) => {
+  const columns = table.getAllColumns()
+  const [localOrder, setLocalOrder] = useState(() => transformToLocalOrder(table))
+
+  const setTableOrder = () => {
+    const newOrder = transformToTableOrder(localOrder)
+    // waiting for the animation to finish so the rerender doesn't cause a visible freeze
+    setTimeout(() => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/crm/CustomizeViewModal.tsx:41]: I wonder if we could use `onAnimationComplete` instead of these timeouts 🤔

BrickheadJohnny (approved):
Left one comment (just an idea), but otherwise looks & feels really good to me!

src/components/[guild]/crm/CustomizeViewModal.tsx
Comment (dovalid) [reply src/components/[guild]/crm/CustomizeViewModal.tsx:44]: Hmm I didn't know about that, I've tried it but it doesn't work on Reorder.Item unfortunately, so I've just extended the comment with info about it, ty!

PR #997, opened 2024-01-02
Use await and loop for polling jobs

Devidxyz:
- If the last GET haven't finished we should not call the GET again.
  - If a GET took more than 3 seconds, we called the API again, creating more requests and more load. Now we wait until the last call is finished then wait 3 seconds.
- Also updated the error fields.
- The "Job not found" part should not happen in normal scenarios, only when the redis is manually flushed (by me).

You can check it here: https://guild-xyz-git-await-job-polling-zgen.vercel.app/our-guild


BrickheadJohnny (approved):
this looks great, but can you center a div?

PR #999, opened 2024-01-04
[GUILD-2074] - Fix GitHub reauth alert

cs-balazs:
[Linear issue](https://linear.app/guildxyz/issue/GUILD-2074/github-guild-creation-fix-wrong-scope-error)

- Since there is no exact way to tell if a reauth is needed, or the user simply doesn't have any repos, this PR modifies the error alert to account for both cases
- There are three minor bugs I've found on the way:
  - The skeleton's `key` values were wrong, I believe they all were undefined, instead of the intended index values
  - When selecting a platform on the create-guild page, the modal didn't open when a platform connection also happened
  - The user SWR state contained invalid data when a reauth happened, because the mutation just appended the platformUser object to the array


BrickheadJohnny (approved):
Awesome, thanks for the fix!

PR #1000, opened 2024-01-05
feat(Secret rewards): validate text(s)' length

gdn992:
[issue](https://linear.app/guildxyz/issue/GUILD-1512/secret-rewards-validate-texts-length)


BrickheadJohnny (approved):
🚀

PR #1001, opened 2024-01-08
feat(Add reward): Add spinner to all AddPlatformPanel

gdn992:
[GUILD-2046](https://linear.app/guildxyz/issue/GUILD-2046/add-loading-fallback-to-addplatformpanels)

### DIFF
--- a/src/platforms/platforms.tsx
+++ b/src/platforms/platforms.tsx
@@ -107,9 +107,9 @@ type PlatformData<
   }
 >
 
-export const AddPlatformPanelLoadingSpinner = () => (
+const AddPlatformPanelLoadingSpinner = () => (
   <Center w="full" h="51vh">
-    <Spinner size="xl" />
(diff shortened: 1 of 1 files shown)


src/platforms/platforms.tsx
```diff
@@ -103,6 +107,12 @@ type PlatformData<
   }
 >
 
+const AddPlatformPanelLoadingSpinner = () => (
+  <Center w="full" h="51vh">
+    <Spinner />
```
BrickheadJohnny, line 112: I think the spinner could be bigger, but otherwise it looks good to me!

gdn992: okay, I'll fix it

gdn992: image
XL maybe better than L because the content size quite big.

BrickheadJohnny (approved):
🚀

PR #1003, opened 2024-01-09
feat: new Twitter follow/like/retweet requirements

BrickheadJohnny:
Added 3 new Twitter requirement types:
- `TWITTER_FOLLOW_V2`
- `TWITTER_LIKE_V2`
- `TWITTER_RETWEET_V2`

The API rate limits are quite low, so we decided to use [Web Intents](https://developer.twitter.com/en/docs/twitter-for-websites/web-intents/overview). The main logic for these new types is inside the `TwitterIntent` component.


src/requirements/Twitter/components/TwitterIntent.tsx
```diff
@@ -0,0 +1,133 @@
...
+    onSuccess: () => {
+      mutateAccess()
+      setHasClicked(false)
+    },
+  })
+
+  const [hasClicked, setHasClicked] = useState(false)
+  // Calling the callback endpoint only on refocus
+  useSWR(
+    hasClicked ? ["twitterRequirement", requirementId] : null,
```
cs-balazs, line 82: Not sure, but maybe include the `userId` here as well? I guess it only matters if the user switches to a different account, and clicks again(?)

src/requirements/Twitter/components/TwitterIntent.tsx
```diff
@@ -0,0 +1,133 @@
...
+      revalidateOnMount: false,
+      revalidateIfStale: false,
+      revalidateOnReconnect: false,
+      revalidateOnFocus: true,
+      refreshInterval: 0,
+    }
+  )
+
+  const onClick = () => {
+    setHasClicked(true)
```
cs-balazs, line 101: Do we trigger a `completeAction` call here in all cases? I mean, even if the user has access. If so, is it intended?

cs-balazs (approved):
Great changes!
Approving, as my comments are just minor questions  🚀

src/requirements/Twitter/components/TwitterIntent.tsx
BrickheadJohnny: yup, but that's why I check `hasAccess` in line 84 ;)

src/requirements/Twitter/components/TwitterIntent.tsx
BrickheadJohnny: Oh, good point, will make sure to add it here!

src/requirements/Twitter/components/TwitterIntent.tsx
cs-balazs: That makes sense, thank you!

PR #1004, opened 2024-01-10
feat: NFT collect page edit description

gdn992:
[GUILD-1817](https://linear.app/guildxyz/issue/GUILD-1817/nft-collect-page-add-edit-description-ui)


src/components/[guild]/RoleCard/components/EditRewardDescription/EditRewardModalContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+      toast({
+        status: "success",
+        title: "Successfully updated reward description",
+      })
+      onClose()
+    },
+  })
+
+  return (
+    <FormProvider {...methods}>
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RoleCard/components/EditRewardDescription/EditRewardModalContent.tsx:44]: I think we don't really need a provider here, since don't use the `useFormContext` hook inside the child components

src/pages/[guild]/collect/[chain]/[address].tsx
```diff
@@ -175,13 +182,37 @@ const Page = ({
                   <CollectibleImage src={image} isLoading={isLoading} />
 
                   <Stack spacing={6}>
-                    <Heading
-                      as="h2"
-                      fontFamily="display"
-                      fontSize={{ base: "3xl", lg: "4xl" }}
+                    <HStack
+                      flex="1 0 auto"
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/collect/[chain]/[address].tsx:186]: Don't we just need `justifyContent="space-between"` here? 👀

src/pages/[guild]/collect/[chain]/[address].tsx
```diff
@@ -55,6 +61,7 @@ const Page = ({
   chain: chainFromProps,
   address: addressFromProps,
 }: Omit<Props, "fallback">) => {
+  const { isOpen, onClose, onOpen } = useDisclosure() // for EditRewardDescriptionModal
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/collect/[chain]/[address].tsx:64]: Let's remove the comment and rename the variables instead, so it's easier to find them in the code, like this:
```ts
const {
  isOpen: isEditRewardDescriptionModalOpen,
  ...
} = useDisclosure()
```

src/pages/[guild]/collect/[chain]/[address].tsx
```diff
@@ -175,13 +182,37 @@ const Page = ({
...
 
                   <Stack spacing={6}>
-                    <Heading
-                      as="h2"
-                      fontFamily="display"
-                      fontSize={{ base: "3xl", lg: "4xl" }}
+                    <HStack
+                      flex="1 0 auto"
+                      justifyContent="flex-end"
+                      animation="slideFadeIn .2s"
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/collect/[chain]/[address].tsx:188]: Looks cool, but since the most important part on the page is the NFT image, I wouldn't add an extra animation for the title

BrickheadJohnny (changes requested):

src/platforms/ContractCall/ContractCallCardMenu.tsx
```diff
@@ -1,16 +1,46 @@
...
 
-const ContractCallCardMenu = ({ platformGuildId }: Props): JSX.Element => (
-  <PlatformCardMenu>
-    <EditRewardAvailibilityMenuItem platformGuildId={platformGuildId} />
-    <RemovePlatformMenuItem platformGuildId={platformGuildId} />
-  </PlatformCardMenu>
-)
+const ContractCallCardMenu = ({ platformGuildId }: Props): JSX.Element => {
+  const { guildPlatforms } = useGuild()
+  const { isAdmin } = useGuildPermission()
```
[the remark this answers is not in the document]
gdn992: Ohh, okay, I'll fix it

src/pages/[guild]/collect/[chain]/[address].tsx
```diff
@@ -175,13 +182,37 @@ const Page = ({
...
+                        {name}
+                      </Heading>
+                      <Spacer m="0" />
+                      {isAdmin && (
+                        <>
+                          <IconButton
+                            icon={<Icon as={PencilSimple} />}
+                            size="sm"
+                            rounded="full"
+                            aria-label="Edit role"
```
Comment (dovalid) [inline src/pages/[guild]/collect/[chain]/[address].tsx:204]: leftover from copypasta, should be 'Edit description'

src/pages/[guild]/collect/[chain]/[address].tsx
Comment (gdn992) [reply src/pages/[guild]/collect/[chain]/[address].tsx:186]: Fixed

src/pages/[guild]/collect/[chain]/[address].tsx
Comment (gdn992) [reply src/pages/[guild]/collect/[chain]/[address].tsx:64]: Yes I thought about that too

src/pages/[guild]/collect/[chain]/[address].tsx
Comment (gdn992) [reply src/pages/[guild]/collect/[chain]/[address].tsx:188]: I removed it

src/components/[guild]/RoleCard/components/EditRewardDescription/EditRewardDescriptionModal.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const EditRewardDescriptionModal: React.FC<Props> = ({
+  guildPlatform,
+  isOpen,
+  onClose,
+}: Props): JSX.Element => (
+  <Modal isOpen={isOpen} onClose={onClose} size={"xl"}>
+    <ModalOverlay />
+    <ModalContent>
+      <EditRewardModalContent guildPlatform={guildPlatform} onClose={onClose} />
```
Comment (dovalid) [inline src/components/[guild]/RoleCard/components/EditRewardDescription/EditRewardDescriptionModal.tsx:19]: Why is it separated into another component? They can't be used separately and the files aren't too big, I think we should keep this logic in one file

Comment (gdn992) [reply src/components/[guild]/RoleCard/components/EditRewardDescription/EditRewardDescriptionModal.tsx:19]: I thought it is good if I don't load the forms logic just after the modal is opened. Like this it doesn't need to reset it.

Comment (dovalid) [reply src/components/[guild]/RoleCard/components/EditRewardDescription/EditRewardDescriptionModal.tsx:19]: I see, makes sense! In this case, we should keep the two components, but move the other one in this file too, so the logic stays batched. + would be nice to add a comment about it

src/pages/[guild]/collect/[chain]/[address].tsx
```diff
@@ -182,12 +193,32 @@ const Page = ({
...
+                        {isAdmin && (
+                          <>
+                            <IconButton
+                              icon={<Icon as={PencilSimple} />}
+                              size="sm"
+                              rounded="full"
+                              aria-label="Edit description"
+                              onClick={onOpen}
+                            />
+                            <EditNFTDescriptionModal
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/collect/[chain]/[address].tsx:214]: Since this component is only loaded for admins, I'd suggest you to load it dynamically only on client-side with `next/dynamic`, you can find examples in our codebase, but let me know if you have any questions!
You could event create an `EditNFTDescriptionButton` component with the `useDisclosure` hook (in that case you won't need to rename the isOpen, onOpen, onClose props!) + the `IconButton` &  `EditNFTDescriptionModal`

BrickheadJohnny (approved):
🚀

PR #1011, opened 2024-01-16
feat: Show text rewards after claim period is over

gdn992:


src/hooks/useGetUserRewards.ts
```diff
@@ -0,0 +1,24 @@
+import { useFetcherWithSign } from "../utils/fetcher"
+import useSWR from "swr"
+import useGuild from "../components/[guild]/hooks/useGuild"
+import useUser from "../components/[guild]/hooks/useUser"
+
+export const useGetUserRewards = () => {
```
BrickheadJohnny, line 6: nit: we could simplify the name, it could be `useUsersRewards`

src/hooks/useGetUserRewards.ts
```diff
@@ -0,0 +1,24 @@
...
+  const { id: guildId } = useGuild()
+  const { id: userId } = useUser()
+
+  const fetcherWithSign = useFetcherWithSign()
+  const fetcher = (key: string) =>
```
BrickheadJohnny, line 11: fetcher is a pretty general name (in fact, we have a `fetcher.ts` file in this repo), so I'd suggest you to find another name for this function, e.g. `fetchUsersRewards`

src/hooks/useGetUserRewards.ts
```diff
@@ -0,0 +1,24 @@
...
+    fetcherWithSign([
+      key,
+      {
+        method: "GET",
+        body: {},
+      },
+    ])
+
+  return useSWR<any[]>(
```
BrickheadJohnny, line 20: Could you please add proper types here?

src/hooks/useIsRewardClaimed.ts
```diff
@@ -0,0 +1,12 @@
+import { RolePlatform } from "../types"
+import { useGetUserRewards } from "./useGetUserRewards"
+
+export const useIsRewardClaimed = (platformId: RolePlatform["id"]) => {
```
BrickheadJohnny, line 4: `platformId` is a little bit misleading here in my opinion, since `platformId` is the ID of a specific platform type (e.g. Discord or Telegram), and not a reward's ID

src/hooks/useIsRewardClaimed.ts
```diff
@@ -0,0 +1,12 @@
...
+  const { isLoading, data: userRewards } = useGetUserRewards()
+
+  const claimed = !isLoading
```
BrickheadJohnny, line 7: I think we don't need this ternary here, this expression will return either a reward or `undefined` anyways

src/platforms/SecretText/TextCardButton.tsx
```diff
@@ -35,6 +38,8 @@ const TextCardButton = ({ platform }: Props) => {
     typeof rolePlatform?.capacity === "number" &&
     rolePlatform?.capacity === rolePlatform?.claimedCount
       ? "All available rewards have already been claimed"
+      : claimed
```
BrickheadJohnny, line 41: If the user has claimed the reward, we shouldn't display the tooltip, so we don't need these two lines here

src/platforms/components/TextReward.tsx
```diff
@@ -58,11 +61,13 @@ const SecretTextReward = ({ platform, withMotionImg }: RewardProps) => {
           tooltipLabel:
             platform?.capacity === platform?.claimedCount
               ? "All available rewards have already been claimed"
+              : claimed
```
BrickheadJohnny, line 64: Same here - for the tooltip, you can use `isDisabled` ;)

BrickheadJohnny (changes requested):

src/platforms/components/TextReward.tsx
gdn992: Hmm, this button is a link button and it's label is not tell user what they can do, and it's Tooltip is not conditionally rendered.
So I'm thinking, maybe I'll put "You can reveal the secret" text back to here.

gdn992: Hmm, i figured out another solution, I'll put the claimed variable into the if statement. I tested out and it is working perfectly

gdn992: when it's claimed we don't need to disabled the button, and we can reveal the secret

BrickheadJohnny (approved):

BrickheadJohnny:
@gdn992 feel free to merge this when you're ready 😉

PR #1014, opened 2024-01-17
feat(useAccess): caching

BrickheadJohnny:
TODO: refactor the hook, so it only returns an array of access check results (even for per-role access check)


BrickheadJohnny:
We've figured out a better solution for this.

PR #1021, opened 2024-01-22
fix(GuildTabs): debug&fix "/undefined" page

gdn992:
The navigation home tab from events tab when fetching the events was failed or not done causes navigation to the guild.xyz/undefined guild page.


BrickheadJohnny (approved):
lgtm, thanks! 🚀

PR #1022, opened 2024-01-22
feat: show points card in access hub

FBalint:


src/platforms/Points/TextCardButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+const PointsCardButton = ({ platform }: Props) => {
+  const { urlName } = useGuild()
+  const id = platform.id
+
+  return (
+    <>
+      <Link href={`/${urlName}/leaderboard/${id}`}>
+        <Button onClick={() => {}} w="full">
```
BrickheadJohnny, line 19: I think we can use a `LinkButton` here instead of a `Link` and a `Button`

src/platforms/Points/EditPointsModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+      imageUrl: currentImage,
+    },
+  })
+
+  const { control } = methods
+  const localName = useWatch({ control, name: "name" })
+  const localImage = useWatch({ control, name: "imageUrl" })
+
+  useEffect(() => {
+    if (isOpen) methods.reset({ name: currentName, imageUrl: currentImage })
```
BrickheadJohnny, line 63: Consider passing this to the `onClose` prop of the modal, that way we don't need to check if `isOpen` is true, instead we'll just reset the form when the user closes the modal:

```tsx
<Modal isOpen={isOpen} onClose={() => {
  onClose()
  methods.reset(...)
}}
...
>
```

src/platforms/Points/PointsRewardCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <>
+      <RewardCard
+        label={platforms.POINTS.name}
+        title={isLoading ? null : text}
+        image={imageUrl || <Star />}
+        colorScheme={platforms.POINTS.colorScheme}
+        cornerButton={
+          isAdmin && (
+            <PointsCardMenu
```
BrickheadJohnny, line 44: Let's import this component dynamically with `next/dynamic` & `ssr: false`, so we don't include it in the bundle for simple users ;)

src/platforms/Points/PointsRewardCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const info =
+    rank !== undefined ? `${numberToOrdinal(rank)} on the leaderboard` : ``
+
+  return (
+    <>
+      <RewardCard
+        label={platforms.POINTS.name}
+        title={isLoading ? null : text}
+        image={imageUrl || <Star />}
```
BrickheadJohnny, line 40: I think we could wrap this `Star` icon in a `Circle`, you can find a similar solution in the `useSecretTextCardProps` hook

BrickheadJohnny (approved):
Left a couple of comments, but I like this overall! 🚀

src/platforms/Points/PointsRewardCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <>
+      <RewardCard
+        label={platforms.POINTS.name}
+        title={isLoading ? null : text}
+        image={
+          imageUrl || (
+            <Circle size={10} bgColor={bgColor}>
+              <Star color="white" />
```
dovalid, line 50: We should use our own Star icon that we use everywhere else for points

src/platforms/Points/PointsRewardCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+        cornerButton={
+          isAdmin && (
+            <DynamicPointsCardMenu
+              platformGuildId={guildPlatform.platformGuildId}
+            ></DynamicPointsCardMenu>
+          )
+        }
+        description={info}
+      >
+        <PointsCardButton platform={guildPlatform}></PointsCardButton>
```
dovalid, line 64: We could use self-closed JSX here

src/platforms/Points/TextCardButton.tsx
```diff
@@ -0,0 +1,29 @@
+import { Icon } from "@chakra-ui/react"
```
dovalid, line 1: The name of the file remained TextCardButton, should be PointsCardButton

src/platforms/Points/TextCardButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+const PointsCardButton = ({ platform }: Props) => {
+  const { urlName } = useGuild()
+  const id = platform.id
+
+  return (
+    <>
+      <LinkButton
+        colorScheme={"gray"}
+        href={`/${urlName}/leaderboard/${id}`}
+        onClick={() => {}}
```
dovalid, line 20: This onClick doesn't do anything

src/platforms/Points/EditPointsModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <ModalHeader>Edit points</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <FormProvider {...methods}>
+            <Stack spacing={8}>
+              <AddNewPointsType
+                name={name}
+                imageUrl={imageUrl}
+                isOptional={false}
+              ></AddNewPointsType>
```
dovalid, line 82: We could use self-closed JSX here

src/platforms/Points/EditPointsModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+              <Button
+                colorScheme="indigo"
```
dovalid, line 85: This should be green (will give more context sync)

PR #1025, opened 2024-01-23
[GUILD-2148] - Join status indication

cs-balazs:
# Join status indication

## Main changes
- `JoinStepIndicator` as a common component for the indicator, used for all steps
   - `JoinStatusStepIndicator` is a more specific version of this, which takes a join state object, and returns an appropriate `JoinStepIndicator`
- `ProgressJoinStep` is responsible for correctly indicating the current join progress. Takes an entity, as the indicated data is relative to an entity
   - This also wraps `SatisfyRequirementsJoinStep`
- `JoinStateCount` for `x/y`-like subtitles
- `mapAccessJobState` util for mapping a raw join job to a more useable state object, keeping only the data, which we are going to display
- `getJoinStepIndicatorProps` util maps a join progress object to appropriate input for `JoinStepIndicator`

## Differences from Figma design
- The current implementation displays _checked_ requirements instead of the _satisfied_ ones. This seemed like a good idea, since this way, the progress circle does a complete circle, displaying the satisfied ones didn't seem to provide a clear progression feedback
  - Lmk what you think about this! 
- The estimation for Discord is missing, afaik this estimation is not currently returned by the core
  - Pinging @Devidxyz here as I'm not sure if this is really the case


Devidxyz:
> The estimation for Discord is missing, afaik this estimation is not currently returned by the core
> * Pinging @Devidxyz here as I'm not sure if this is really the case
It's possible, but it's easy to add. (Assuming we are talking about the delayed state, which is rare.) Anyway I think we don't need to do everything in one PR. The current state is amazing ✨ 

One thing I noticed 2/2 times: after the last step the 0th step (green join button) is showing for half a second, it's not that annoying, just noticed.

Furthermore some incremental updates I can imagine in the future:
- handling the delay state
- if i close the popup I see which requirements are already checked, can be very useful when some take really long time
- the access check / re-check status (which is basically a join) also shows the progress

cs-balazs:
Thank you @Devidxyz for checking the PR out!

> One thing I noticed 2/2 times: after the last step the 0th step (modal with the green join button) is showing for half a second, it's not that annoying, just noticed.

Nice catch! Fixed in https://github.com/guildxyz/guild.xyz/pull/1025/commits/99a5a1cfe8b57da6a5435bc6d6f27fd767c7b50c!


I've also added a bit of delay to the final state, because in some cases, we skip a couple of states, and transition immediately to a final state. With a small delay, it seems cleaner, giving some time to see what happened

BrickheadJohnny (approved):
We talked IRL about 1-2 small things which you'll add, so approving it!

dovalid (approved):

PR #1026, opened 2024-01-23
fix: remove default discord reward when creating role

FBalint:


src/components/[guild]/AddAndOrderRoles/components/AddRoleDrawer.tsx
```diff
@@ -53,32 +51,8 @@ const AddRoleDrawer = ({ isOpen, onClose, finalFocusRef }): JSX.Element => {
       },
     })
 
-  const defaultValues: RoleToCreate = {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/AddAndOrderRoles/components/AddRoleDrawer.tsx:56]: I think we use both uncontrolled and controlled fields deep inside our form, so let's keep the default values please & just remove the `discordPlatform ? ...` condition ;) (and set an empty array as default value)

BrickheadJohnny (commented):
Left 1 comment, let's fix that and then feel free to merge this PR, thanks! 🚀

src/components/[guild]/AddAndOrderRoles/components/AddRoleDrawer.tsx
```diff
@@ -62,23 +62,11 @@ const AddRoleDrawer = ({ isOpen, onClose, finalFocusRef }): JSX.Element => {
...
-            visibility: Visibility.PUBLIC,
-          },
-        ]
-      : [],
+    rolePlatforms: [],
   }
 
   const methods = useForm({
     mode: "all",
-    defaultValues,
```
Comment (BrickheadJohnny) [inline src/components/[guild]/AddAndOrderRoles/components/AddRoleDrawer.tsx:81]: Let's keep this too ;)

BrickheadJohnny (approved):

PR #1030, opened 2024-01-25
feat(TwitterRequirements): add-4-new-twitter-requirements

gdn992:
I had to solve some other things also, 
the order in the requirement select was changed
the twitter texts changed to X, and also the icon was replaced from bird to X

https://vercel.com/zgen/guild-xyz/HN4YwhUVKgkWgy5CDnqFBFU3RadL


src/requirements/requirements.ts
```diff
@@ -20,6 +20,7 @@ export const REQUIREMENTS_DATA = [
   {
     icon: Wallet,
     name: "Free",
+    disabled: false,
```
BrickheadJohnny, line 23: Oh, okay hehe, this was out of scope, but I appreciate it. Let's call is `isDisabled` please (just for consistency)

src/requirements/Twitter/components/TwitterVerificationSelect.tsx
```diff
@@ -0,0 +1,54 @@
...
+    label: "Business",
+    value: "business",
+  },
+  {
+    label: "Government",
+    value: "government",
+  },
+]
+
+const TwitterVerificationSelect = ({
```
BrickheadJohnny, line 29: Since we use this component only for 1 requirement, let's mirror the requirement's name here and rename it to `TwitterAccountVerified`

public/requirementLogos/X.svg
BrickheadJohnny, line 1: This is not used anywhere, so you can remove it

public/platforms/x.png
BrickheadJohnny, line 1: Let's make a round SVG icon with black background and white 𝕏 symbol on it. I can help you in that if you wish! And please use the new X logo as the requirement's image in `TwitterRequirement` too ;)

src/requirements/Twitter/components/TwitterVerificationSelect.tsx
```diff
@@ -0,0 +1,54 @@
...
+    setValue,
+    formState: { errors },
+  } = useFormContext()
+  const defaultValue: ValueType = "any"
+
+  return (
+    <FormControl isInvalid={!!parseFromObject(errors, baseFieldPath)?.type}>
+      <FormLabel>Verification type</FormLabel>
+
+      <ControlledSelect
```
BrickheadJohnny, line 42: You can use this component as you'd use a react-hook-form `Controller`, so you just need to pass a name, defaultValue, options, and no need to handle onChange ;) 

I'm pretty sure you won't event need to handle that "any" value, you can just set the first option's value to `null` and it'll work out of the box!

BrickheadJohnny (changes requested):

src/requirements/Twitter/components/TwitterVerificationSelect.tsx
gdn992: image
I've got this message when I changed it to null, that's because I chosen a more readable value.

gdn992: I change it to null, and add a todo somewhere

BrickheadJohnny (approved):
Approved, but please double-check the copy with Nasheq / Dovalid before merging ;)

src/requirements/Twitter/TwitterRequirement.tsx
```diff
@@ -58,33 +62,77 @@ const TwitterRequirement = (props: RequirementProps) => {
...
               </>
             )
           case "TWITTER_FOLLOWER_COUNT":
-            return `Have at least ${Math.floor(
-              requirement.data.minAmount
-            )} followers on Twitter`
+            return (
+              <Text as="span">
+                {"Have at least "}
+                <DataBlock>{pluralize(minCount, "follower")}</DataBlock>
```
dovalid, line 88: I wouldn't put these into DataBlocks, the information is simple enough here to work better as plain text imo

src/requirements/Twitter/TwitterRequirement.tsx
```diff
@@ -17,6 +17,10 @@ import TwitterListLink from "./components/TwitterListLink"
 import TwitterTweetLink from "./components/TwitterTweetLink"
 import TwitterUserLink from "./components/TwitterUserLink"
 
+// todo if it's working well I can change the original pluralize function
```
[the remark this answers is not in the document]
gdn992: It was used because I used some irregular plural words that was not covered by the original pluralize function, but I created a perfect one :) (not this)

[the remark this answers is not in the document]
gdn992: I removed it

dovalid (approved):

PR #1031, opened 2024-01-25
feat: create superadmin activity log page

FBalint:
### Whats new
- new `superadmin/activity` url
- show 404 page if user is not superadmin
- in `ActivityLogContext`,  the `isUserActivity` prop is changed to `activityLogType` ("all",  "guild", and "user" activity logs)
- suggestions are rendered according to the `activityLogType`, moved to a separate `Suggestions` component
- fix for reward filter tags - after selecting a suggestion, they were not visible in the filter; now they are
---
### Todo
- [x] Test with a superadmin account
- [x] Clear all filters button is not working

### DIFF
--- a/src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx
+++ b/src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx
@@ -328,7 +328,6 @@ const FilterTag = ({
                 case "roleId":
                   return (
                     <RoleSuggestions
-                      guildId={guildId}
                       inputValue={inputValue}
                       getOptionProps={getOptionProps}
                     />
--- a/src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx
+++ b/src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx
@@ -307,53 +307,53 @@ const FilterTag = ({
         )}
       </Tag>
 
-      {["guildId", "roleId", "rolePlatformId", "action"].includes(filter) && (
-        <Dropdown
-          {...{
-            ...positionerProps,
-            style: { ...positionerProps.style, ...positionerStyle },
-          }}
-        >
-          <Stack spacing={0} {...contentProps}>
-            {(() => {
-              switch (filter) {
-                case "guildId":
-                  if (isSuperadminActivityLog) return <></>
-                  return (
-                    <GuildSuggestions
-                      inputValue={inputValue}
-                      getOptionProps={getOptionProps}
-                    />
-                  )
-                case "roleId":
-                  return (
-                    <RoleSuggestions
-                      inputValue={inputValue}
-                      getOptionProps={getOptionProps}
-                    />
-                  )
-                case "rolePlatformId":
-                  return (
-                    <RewardSuggestions
-                      guildId={guildId}
-                      inputValue={inputValue}
-                      getOptionProps={getOptionProps}
-                    />
-                  )
-                case "action":
-                  return (
-                    <ActionSuggestons
-                      inputValue={inputValue}
-                      getOptionProps={getOptionProps}
-                    />
-                  )
-                default:
-                  return null
-              }
-            })()}
-          </Stack>
-        </Dropdown>
-      )}
+      {["guildId", "roleId", "rolePlatformId", "action"].includes(filter) &&
+        !(isSuperadminActivityLog && filter == "guildId") && (
+          <Dropdown
+            {...{
+              ...positionerProps,
+              style: { ...positionerProps.style, ...positionerStyle },
+            }}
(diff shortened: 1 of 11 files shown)


src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx
```diff
@@ -50,10 +52,19 @@ const FilterTag = ({
...
   const closeBtnFocusBgColor = useColorModeValue("blackAlpha.300", "whiteAlpha.500")
 
-  const { getFilter, updateFilter, removeFilter } = useActivityLogFilters()
+  const { activeFilters, getFilter, updateFilter, removeFilter } =
+    useActivityLogFilters()
+
+  const { activityLogType } = useActivityLog()
+  const isSuperadminActivityLog = activityLogType === "all"
+
+  const filterGuildId = isSuperadminActivityLog
```
Comment (BrickheadJohnny) [inline src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx:61]: I think we could use `getFilter("guildId")` here, right? 👀

src/components/[guild]/activity/ActivityLogFiltersBar/components/Suggestions.tsx
```diff
@@ -0,0 +1,122 @@
+import { Stack, Text } from "@chakra-ui/react"
+import * as combobox from "@zag-js/combobox"
+import { HTMLAttributes, useMemo } from "react"
+import { ActivityLogType, useActivityLog } from "../../ActivityLogContext"
+import { Filter, SupportedQueryParam } from "./ActivityLogFiltersContext"
+import Suggestion from "./Suggestion"
+
+interface Props {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/activity/ActivityLogFiltersBar/components/Suggestions.tsx:8]: let's just stick to type here, since we usually use types

src/components/[guild]/activity/ActivityLogFiltersBar/components/Suggestions.tsx
```diff
@@ -0,0 +1,122 @@
...
+  contentProps: HTMLAttributes<HTMLElement>
+  activityLogType: ActivityLogType
+  activeFilters: Filter[]
```
Comment (BrickheadJohnny) [inline src/components/[guild]/activity/ActivityLogFiltersBar/components/Suggestions.tsx:11]: We could omit this prop & use it from the `useActivityLogFilters` hook instead

src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/components/RoleSuggestions.tsx
```diff
@@ -6,12 +6,17 @@ import Suggestion from "../../Suggestion"
...
 type Props = {
+  guildId?: string | number
   inputValue?: string
   getOptionProps: (props: combobox.OptionProps) => HTMLAttributes<HTMLElement>
 }
 
-const RoleSuggestions = ({ inputValue, getOptionProps }: Props): JSX.Element => {
-  const { id: guildId, roles } = useGuild()
+const RoleSuggestions = ({
+  guildId,
```
Comment (BrickheadJohnny) [inline src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/components/RoleSuggestions.tsx:15]: Do we actually need to pass the `guildId` prop here (and in the `RewardSuggestions` component too), or could we use it from context (e.g. active filters context)?

BrickheadJohnny (commented):
Great work, I really like that new `Suggestions` component, that cleaned up `FiltersInput` nicely! Left a couple of suggestions & 1 question ;)

src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx
Comment (FBalint) [reply src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx:61]: `getFilter` takes the ID of the filter, which is a long string like `cd4bfcf1-ba06-46a2-9515-840476a4024d`, not the query string
should we add a `getFilterByQueryString` function to the hook?

src/components/[guild]/activity/ActivityLogFiltersBar/components/Suggestions.tsx
Comment (FBalint) [reply src/components/[guild]/activity/ActivityLogFiltersBar/components/Suggestions.tsx:11]: nice catch, thanks!

src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/components/RoleSuggestions.tsx
Comment (FBalint) [reply src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/components/RoleSuggestions.tsx:15]: Currenlty, both of these components are used only in the `FilterTag` component, which can directly pass them the `guildId` as a prop. Why would context be preferable in this case? Is there a convention that I should follow when deliberating the usage of props over context?

src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/activity/ActivityLogFiltersBar/components/FilterTag/FilterTag.tsx:61]: Oh, sorry, I completely missed that, then we can leave it as is!

BrickheadJohnny (approved):

PR #1034, opened 2024-01-26
refactor: useUsersGuildPins hook

FBalint:


src/hooks/useUsersGuildPins.ts
```diff
@@ -8,7 +8,89 @@ import {
...
+      ? await client.multicall({
+          contracts: contracts,
+        })
+      : []
+
+  const errors: Error[] = results
+    .filter((result) => result.status === "failure")
+    .map((result) => result.error)
+
+  const pinIds: bigint[] = results
```
BrickheadJohnny, line 37: This type is redundant, it'll have the proper type since you're already casting result as bigint ;)

src/hooks/useUsersGuildPins.ts
```diff
@@ -8,7 +8,89 @@ import {
...
+  }))
+
+  const results =
+    contractCalls.length > 0
+      ? await client.multicall({
+          contracts: contractCalls,
+        })
+      : []
+
+  const tokenURIs = results.map((result, index) => ({
```
BrickheadJohnny, line 63: This variable name is a bit misleading, `tokenURI` is a base64 encoded metadata, this variable contains that base64 string, the chainId, and the tokenId too, so let's find a better name for it please

src/hooks/useUsersGuildPins.ts
```diff
@@ -8,7 +8,89 @@ import {
...
+  }))
+
+  const errors: Error[] = results
+    .filter((res) => res.status === "failure")
+    .map((res) => res.error)
+
+  return { tokenURIs, errors }
+}
+
+const tokenURItoMetadataJSON = (tokenURI: {
```
BrickheadJohnny, line 76: This is also misleading for the same reason, but we could call it `getMetadataJSON` or something like that, lmk if you have a better name in your mind

src/hooks/useUsersGuildPins.ts
```diff
@@ -19,70 +101,48 @@ const fetchGuildPinsOnChain = async (
...
       address: GUILD_PIN_CONTRACTS[chain].address,
-      functionName: "tokenOfOwnerByIndex",
-      args: [address, BigInt(i)],
+      functionName: "balanceOf",
+      args: [address],
     })
-
-    if (newTokenId) usersGuildPinIdsOnChain.push(newTokenId)
+  } catch (e) {
+    return { usersPinsMetadataJSONs: [], errors: [{ error: e, status: "failure" }] }
```
BrickheadJohnny, line 113: Please type this (even just by adding type casts), so we get a nice clean return type for the `fetchGuildPinsOnChain` function!

src/hooks/useUsersGuildPins.ts
```diff
@@ -91,20 +151,40 @@ const fetchGuildPins = async ([_, addresses]: [string, User["addresses"]]) => {
     )
   )
 
-  return responseArray.flat()
+  const { allUsersPins, allErrors } = responseArray.reduce(
```
BrickheadJohnny, line 154: This reduce unfortunately returns `any[]`s, so let's add a type to it ;)

BrickheadJohnny (changes requested):
Great refactor, thank you! Added a couple of comments, I think we could make this code a little bit easier to understand with some minor changes 👍

FBalint:
> Great refactor, thank you! Added a couple of comments, I think we could make this code a little bit easier to understand with some minor changes 👍

Thanks! I updated the typing, the naming, and fixed an error handling issue. Let me know what you think!

src/hooks/useUsersGuildPins.ts
```diff
@@ -8,7 +8,101 @@ import {
...
+    .filter((res) => res.status === "failure")
+    .map((res) => res.error)
+
+  return { tokenInfo, errors }
+}
+
+type TokenWithMetadata = {
+  chainId: Chains
+  tokenId: number
+  image: string
```
BrickheadJohnny, line 85: I think image is already defined in `GuildPinMetadata` ;)

src/hooks/useUsersGuildPins.ts
```diff
@@ -91,20 +181,40 @@ const fetchGuildPins = async ([_, addresses]: [string, User["addresses"]]) => {
...
 
   const evmAddresses = addresses?.filter((address) => address.walletType === "EVM")
 
   const shouldFetch = Boolean(!disabled && isWeb3Connected && evmAddresses?.length)
 
-  return useSWRImmutable<
-    ({ chainId: number; tokenId: number } & GuildPinMetadata)[]
-  >(shouldFetch ? ["guildPins", evmAddresses] : null, fetchGuildPins)
+  const swrData = useSWRImmutable(
+    shouldFetch ? ["guildPins", addresses, includeTestnets] : null,
```
BrickheadJohnny, line 209: We should pass `evmAddresses` here

src/hooks/useUsersGuildPins.ts
```diff
@@ -91,20 +181,40 @@ const fetchGuildPins = async ([_, addresses]: [string, User["addresses"]]) => {
...
+    fetchGuildPins
+  )
+
+  return {
+    ...swrData,
+    data: swrData.data?.usersPins,
+    pinFetchErrors: swrData.data?.errors,
```
BrickheadJohnny, line 216: I think we could just leave this prop out and instead return a boolean in the `error` prop (`swrData.data?.errors?.length > 0`), since we won't use the actual errors anywhere

src/hooks/useUsersGuildPins.ts
FBalint: Nice, do you have a tip for noticing these typing opportunities?

BrickheadJohnny: I'll be honest, I noticed by chance :D
But, I guess you'll just get to know the codebase better over time and you'll notice these too ;)

BrickheadJohnny (approved):
🚀

PR #1037, opened 2024-01-27
fix: disable type field when req form is in edit mode

FBalint:


src/requirements/CoinbaseEAS/CoinbaseEASForm.tsx
```diff
@@ -86,6 +90,7 @@ const CoinbaseEASForm = ({ baseFieldPath }: RequirementFormProps): JSX.Element =
 
             setValue(`${baseFieldPath}.data.val`, undefined)
           }}
+          isDisabled={isEditMode}
```
BrickheadJohnny, line 93: I think this should not be disabled ;)

BrickheadJohnny (commented):
I noticed 1 small issue, but otherwise it looks good to me 🚀 

We'll need to find a solution for requirement types which have multiple subtypes, but we set them automatically (e.g. NFT requirement has 3 subtypes, Token requirement has 2 subtypes). Anyway, we'll solve this issue in another PR!

PR #1038, opened 2024-01-27
refactor: moving role platform status logic to utils

FBalint:


src/utils/rolePlatformHelpers.ts
```diff
@@ -0,0 +1,41 @@
...
+  const endTimeDiff = getTimeDiff(rolePlatform?.endTime)
+
+  const inActiveTimeframe =
+    (startTimeDiff > 0 ||
+      endTimeDiff < 0 ||
+      (typeof rolePlatform?.capacity === "number" &&
+        rolePlatform?.capacity === rolePlatform?.claimedCount)) &&
+    additionalCondition
+
+  return { inActiveTimeframe, startTimeDiff, endTimeDiff }
```
BrickheadJohnny, line 38: `inActiveTimeframe` is a bit confusing, since this variable is `true` if the reward is outside the defined timeframe, let's try another name here ;)

BrickheadJohnny (commented):
I really like this refactor, thanks!
Pushed one more commit, please check it & also left one short comment, lmk your opinion on that too!

src/utils/rolePlatformHelpers.ts
FBalint: You are absolutely right, thanks!
I inspected the logic again, and inverted it, so now it returns an `isAvailable` boolean instead. I also decided to remove the `additionalCondition` prop from the hook, as it takes effort to understand how it affects the return values.

BrickheadJohnny (approved):

PR #1040, opened 2024-01-29
fix: returning null instead of undefined

FBalint:
In the "Edit Availability" modal for a reward, unchecking the "Limit claiming time" option and saving the changes does not remove the claiming time limit as expected. Instead, the time limit remains active.


BrickheadJohnny (approved):
Thanks! 🚀

PR #1042, opened 2024-01-30
feat:Add filter to Guild-"have a role" requirement

gdn992:
feat(Role): create RoleWithRangeFilter component

feat(Role): add filter to the options of the selectable roles

feat(GuildRequirement): add label to display the filter

feat(requirements): Add new GUILD_ROLE_RELATIVE requirement

feat(GuildRequirement): Add new HaveRoleRelative Component


BrickheadJohnny (changes requested):
Left a couple of comments, please check them. It would be also great if we'd reset the form when the requirement type changes, please implement that too.

BrickheadJohnny (approved):

PR #1045, opened 2024-02-01
fix: make email required on guild creation

FBalint:


BrickheadJohnny (approved):
I like how you experimented with the first e-mail field, but I felt like we should keep the flow as consistent as possible, so I refactored it a bit to match the old flow ;) Let me know if this makes sense to you.

FBalint:
> I like how you experimented with the first e-mail field, but I felt like we should keep the flow as consistent as possible, so I refactored it a bit to match the old flow ;) Let me know if this makes sense to you.

I agree that the code is simpler this way! However, should we allow the additional fields to be left empty? It allows guild creation with empty strings, and later these also show up in the guild settings at the contacts section.

BrickheadJohnny:
> > I like how you experimented with the first e-mail field, but I felt like we should keep the flow as consistent as possible, so I refactored it a bit to match the old flow ;) Let me know if this makes sense to you.
> 
> I agree, this approach simplifies the code! However, it could lead users to wonder why a dropdown is present if it's not interactive, potentially giving them a sense of restricted control. Nonetheless, this might not be a significant concern.
> 
> However, should we allow the additional fields to be left empty? It allows guild creation with empty strings, and later these also show up as empty entries in the guild settings at the contacts section.

Oh, good point, could you please solve this issue too? 👀

PR #1048, opened 2024-02-02
Rework access / membership management

cs-balazs:


src/components/[guild]/JoinModal/components/CompleteCaptchaJoinStep.tsx
```diff
@@ -15,8 +15,10 @@ const CompleteCaptchaJoinStep = (): JSX.Element => {
     ?.filter((req) => req.type === "CAPTCHA")
     .map((req) => req.id)
 
-  const { data: accesses } = useAccess()
-  const requirementAccesses = accesses?.flatMap((access) => access.requirements)
+  const { membership } = useMembership()
+  const requirementAccesses = membership?.roles?.flatMap((role) =>
```
Comment (BrickheadJohnny) [inline src/components/[guild]/JoinModal/components/CompleteCaptchaJoinStep.tsx:19]: is it intentional that we filter only true accesses here (and in every place where we check requirementAccesses)? Maybe it isn't necessary, since we check if `req.access` is true in `isDone` logic

src/components/[guild]/LeaveButton/hooks/useLeaveGuild.ts
```diff
@@ -25,7 +23,7 @@ const useLeaveGuild = () => {
         status: "success",
       })
 
-      mutate((prev) => prev.filter(({ guildId }) => guildId !== id), {
+      mutate(undefined, {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/LeaveButton/hooks/useLeaveGuild.ts:26]: should't we just `mutate()` here?

BrickheadJohnny (commented):
very good, very nice, only left 2 questions ;)

src/components/[guild]/JoinModal/components/CompleteCaptchaJoinStep.tsx
Comment (cs-balazs) [reply src/components/[guild]/JoinModal/components/CompleteCaptchaJoinStep.tsx]: Nice catch! It might not be needed then! Looking at the diff, it indeed seems like, that `requirementAccesses` included all the requirements, not just the satisfied ones
Changed in https://github.com/guildxyz/guild.xyz/pull/1048/commits/ddcf71b4d77662b47692b6b9ebc341f92c6896e9!

src/components/[guild]/LeaveButton/hooks/useLeaveGuild.ts
Comment (cs-balazs) [reply src/components/[guild]/LeaveButton/hooks/useLeaveGuild.ts:26]: I'm not sure, the idea here was, that since we left the guild, we don't have any memberships anymore, so we just delete the data by mutating it to undefined

src/components/[guild]/hooks/useIsMember.ts
```diff
@@ -1,13 +1,11 @@
-import useMemberships from "components/explorer/hooks/useMemberships"
-import useGuild from "components/[guild]/hooks/useGuild"
+import useMembership from "components/explorer/hooks/useMemberships"
 
 const useIsMember = (): boolean => {
```
Comment (dovalid) [inline src/components/[guild]/hooks/useIsMember.ts:3]: This appears to be doing the same as hasAccess in useMembership now, so we could remove this hook and just use that (IMO renamed to isMember, since that's more accurate than just having access)

src/components/[guild]/hooks/useAutoStatusUpdate.ts
```diff
@@ -11,39 +11,43 @@ const useAutoStatusUpdate = () => {
   const { id, mutateGuild } = useGuild()
   const { keyPair } = useUserPublic()
 
-  const { data: accesses, mutate: mutateAccess } = useAccess()
-  const { memberships, mutate: mutateMemberships } = useMemberships()
-
-  const roleMemberships = memberships?.find(
-    (membership) => membership.guildId === id
-  )?.roleIds
+  const { onSubmit: onJoin } = useJoin()
```
Comment (dovalid) [inline src/components/[guild]/hooks/useAutoStatusUpdate.ts:14]: We should just remove this file, no? Membership won't change on it's own so we don't have any data to trigger an automatic update by

src/requirements/Captcha/hooks/useVerifyCaptcha.ts
```diff
@@ -15,7 +15,7 @@ const useVerifyCaptcha = () => {
   const showErrorToast = useShowErrorToast()
   const toast = useToast()
 
-  const { mutate } = useAccess()
+  const { onSubmit: onJoin } = useJoin()
```
dovalid, line 18: I think we should rename useJoin to useMembershipUpdate, and onJoins like this to triggerMembershipUpdate, wdyt?

cs-balazs: You mean to rename the import, and the returned value like:
```ts
import useMembershipUpdate from "...";
const { onSubmit: triggerMembershipUpdate } = useMembershipUpdate()
```

Or to crate a separate hook named `useMembershipUpdate`, and return onSubmit as `triggerMembershipUpdate`?

src/components/[guild]/hooks/useAutoStatusUpdate.ts
Comment (cs-balazs) [reply src/components/[guild]/hooks/useAutoStatusUpdate.ts:14]: Oh, yeah, forgot to mention this in the channel, but this feature probably won't work anymore

cs-balazs:
Also, forgot to mention, but we might want to rename `useMemberships` to `useMembership`. Didn't want to do that for now, so there are diffs on the file to review, but we might want to consider it before merging

src/components/[guild]/hooks/useIsMember.ts
Comment (cs-balazs) [reply src/components/[guild]/hooks/useIsMember.ts]: Good idea, changed in https://github.com/guildxyz/guild.xyz/pull/1048/commits/353d7d0d7e82e12bac9af3e1511b22a530b19d95!

src/components/[guild]/hooks/useAutoStatusUpdate.ts
Comment (cs-balazs) [reply src/components/[guild]/hooks/useAutoStatusUpdate.ts]: Removed in https://github.com/guildxyz/guild.xyz/pull/1048/commits/f9dd48e934dbf9e0b2fb0e6da995b1851ac0096d!

src/requirements/Captcha/hooks/useVerifyCaptcha.ts
dovalid: Sry I mean I'd rename the whole useJoin.tsx file and the exported hook to useMembershipUpdate, since it's not really just join anymore, just wrote the comment here cause it illustrates the motivation well

dovalid: Now that I looked at useJoin there's still a reasonable amount of join specific logic, so ideally we'd separate it into two files, an useMembershipUpdate with the core logic that we'd use everywhere, and an useJoin above that with all the join specific logic that we'd use in JoinModal, and later in other places where we have custom join logic now too: https://discord.com/channels/697041998728659035/1202271479744438303/1202385210264522823

PR #1050, opened 2024-02-03
fix: display line breaks in secret text reward

FBalint:


src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddContractCallPanel/components/CreateNftForm/components/ResetHeadingOnEnterPlugin.tsx
```diff
@@ -0,0 +1,40 @@
...
+import {
+  $createParagraphNode,
+  $getSelection,
+  $isRangeSelection,
+  COMMAND_PRIORITY_LOW,
+  KEY_ENTER_COMMAND,
+} from "lexical"
+import { useEffect } from "react"
+
+function ResetHeadingOnEnterPlugin() {
```
Comment (dovalid) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddContractCallPanel/components/CreateNftForm/components/ResetHeadingOnEnterPlugin.tsx:11]: I don't understand what this does, I'm still seem to be able to press enter / shift+enter in a heading to put a linebreak in it, and it renders well at claiming too. I couldn't notice a difference when I commented it out in RichTextDescriptionEditor. What am I missing?

If we still need it, I think we should put this into components/lexical/plugins like other plugins, and the name could be PreventHeadingLinebreakPlugin (if it's supposed to do that?)

Comment (FBalint) [reply src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddContractCallPanel/components/CreateNftForm/components/ResetHeadingOnEnterPlugin.tsx]: I moved it to the lexical pugins dir, and changed it so it also works when adding line break inside the line as well, not just at the end.

PR #1054, opened 2024-02-06
[GUILD-2151] - Fix Delegate.cash inegration

cs-balazs:
Core PR: https://github.com/guildxyz/guild-backend/pull/1254

- The FE utils have also been migrated to the v2 of Delegate.cash
- The same chains have also been added here
- All the Delegate.cash-related actions now call the new endpoint


BrickheadJohnny (approved):
🚀

PR #1063, opened 2024-02-08
feat: add minAmount field to wallet age requirement form

FBalint:


src/requirements/WalletActivity/components/AlchemyFirstTx.tsx
```diff
@@ -1,14 +1,40 @@
...
+  const [minAmount, maxAmount] = useWatch({
+    name: [
+      `${baseFieldPath}.data.timestamps.minAmount`,
+      `${baseFieldPath}.data.timestamps.maxAmount`,
+    ],
+  })
+
+  return (
+    <>
+      <Text colorScheme="gray" fontSize="sm" mt={-2} pl={3}>
```
BrickheadJohnny, line 16: We actually already have a helper text inside the `BlockNumberFormControl` component, could you please check why it's not showing up? 👀

src/requirements/WalletActivity/components/AlchemyFirstTxRelative.tsx
```diff
@@ -1,17 +1,43 @@
...
+  const [minAmount, maxAmount] = useWatch({
+    name: [
+      `${baseFieldPath}.data.timestamps.minAmount`,
+      `${baseFieldPath}.data.timestamps.maxAmount`,
+    ],
+  })
+
+  return (
+    <>
+      <Text colorScheme="gray" fontSize="sm" mt={-2} pl={3}>
```
BrickheadJohnny, line 18: Same here please 😉

src/requirements/WalletActivity/components/BlockNumberFormControl.tsx
```diff
@@ -75,6 +79,7 @@ const BlockNumberFormControl = ({
     register(`${baseFieldPath}.data.${dataFieldName}`, {
       required:
         !!isRequired && !!shouldFetchBlockNumber && "This field is required.",
+      validate: () => (isInvalid ? invalidText : undefined),
```
BrickheadJohnny, line 82: I think we could use the native HTML `min` & `max` attributes on the date pickers, and just make sure that the user can't pick invalid dates. You can see an example inside the `EditRewardAvailabilityModal` component.

BrickheadJohnny (changes requested):
Great work, left a couple of comments on the error handling & helper text. Also, let's validate the copy with the team ;)

src/requirements/WalletActivity/WalletActivityForm.tsx
```diff
@@ -103,17 +85,7 @@ const WalletActivityForm = ({
...
-        .filter(({ value }) => !COVALENT_EXCLUDED_TYPES.has(value))
-        .map(
-          ({ value, ...rest }) =>
-            ({
-              ...rest,
-              value: value.replace("ALCHEMY_", "COVALENT_"),
-            } as SelectOption)
-        )
-    : walletActivityRequirementTypes
+  const supportedRequirementTypes = walletActivityRequirementTypes
```
BrickheadJohnny, line 88: This variable is actually redundant, let's just use `walletActivityRequirementTypes` everywhere

src/requirements/Github/components/MinMaxAmountFormControls.tsx
```diff
@@ -1,54 +1,99 @@
...
   baseFieldPath: string
-  minAmountLabel: string
-  maxAmountLabel: string
+  minAmountLabel?: string
+  maxAmountLabel?: string
+  type?: string
+  timestamp?: boolean
 }
 
 const MinMaxAmountFormControls = ({
```
BrickheadJohnny, line 15: I think we could do a great refactor here actually!

My idea is to completely get rid of the `MinMaxAmountFormControls` component (since it was GitHub specific before & now when someone is reading our code, it is not obvious at first glance that these are actually time inputs), and make 2 different components inside `src/requirements/common`:
- `RelativeMinMaxTimeFormControls` - contains only the absolute time picker logic
- `AbsoluteMinMaxTimeFormControls` - contains only the relative time picker logic
- both of these could expose a `minTimeLabel` & `maxTimeLabel` prop, and we shouldn't use the `type`, `timestamp` & `baseFieldPath` props, instead we could just pass `minTimeFieldName` & `maxTimeFieldName` (strings) and then use those instead of doing these `timestamp ? "timestamps." : ""` conditions everywhere (for errors, if I'm not missing anything, we could just use `parseFromObject(errors, minTimeFieldName)`) ;)

This way, these components would be completely reusable, even if the field paths change in the future, or if we'll need to use them for completely different fields.

Let me know what do you think about this idea, or if you think I missed something!

BrickheadJohnny (changes requested):
I have a refactor idea 💡

BrickheadJohnny (approved):

PR #1068, opened 2024-02-12
Guild 2140 e2e new create guild flow tests

FBalint:


cypress/e2e/0-guild/0-create-guild.spec.ts
```diff
@@ -2,74 +2,212 @@ before(() => {
...
+      cy.visit("/create-guild")
+      cy.connectWallet()
+      fillCustomizeGuildForm()
+    })
+
+    it("can select role templates", () => {
+      // Selecting the first role template
+      getContinueBtn().should("be.disabled")
+      cy.get('#role-checkbox[data-test^="selected-role-"]').should("not.exist")
+      cy.get('[data-test^="role-"]')
```
BrickheadJohnny, line 125: I think you could extract this selector just like you did with the continue button ;)

src/components/create-guild/ChooseTemplate/components/TemplateCard.tsx
```diff
@@ -241,6 +243,8 @@ const TemplateCard = ({
         >
           {selected ? (
             <Circle
+              id={`role-checkbox`}
```
BrickheadJohnny, line 246: Do we actually need the ID attribute here? (I think the `data-test` might be sufficient, but maybe I miss something)

BrickheadJohnny (approved):
Great work!

PR #1070, opened 2024-02-16
feat: xmtp messaging

gdn992:


src/components/common/Layout/components/Account/components/Notifications/MessagingSection.tsx
```diff
@@ -0,0 +1,144 @@
...
+      >
+        {isWeb3InboxLoading || isCheckingXmtpAccess ? (
+          <SubscriptionPromptSkeleton />
+        ) : !web3InboxSubscription && !hasXmtpAccess ? (
+          <SubscriptionPrompt onClick={onOpen} />
+        ) : (
+          <Messages />
+        )}
+      </NotificationsSection>
+      <Modal isOpen={isOpen} onClose={onClose}>
```
BrickheadJohnny, line 69: Let's try this for button labels:
image

src/components/common/Layout/components/Account/components/Notifications/components/xmtp.ts
```diff
@@ -0,0 +1,142 @@
...
+  const { data: signer, isLoading } = useWalletClient()
+
+  const subscribeXmtp = useCallback(async () => {
+    setIsSubscribingXmtp(true)
+    await Client.create(signer, {
+      persistConversations: false,
+      env: "production",
+    })
+      .then(() => {
+        toast({
```
BrickheadJohnny, line 121: The copy isn't right here, this should be XMTP. And also, let's use the custom `useToast` hook with the customized toast:
image

I also noticed that the Sign button next to the XMTP won't get disabled after successfully enabling the identity.

src/components/common/Layout/components/Account/components/Notifications/components/MessagingContext.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const messagingContext = createContext<
+  ReturnType<typeof useSubscribeXmtp> &
+    ReturnType<typeof useWeb3InboxSubscription> &
+    ReturnType<typeof useXmtpAccessChecking>
+>(undefined)
+
+export const useMessagingContext = () => useContext(messagingContext)
+
+export const MessagingWrapper = ({ children }) => {
```
BrickheadJohnny, line 13: Please follow the naming conventions from our other contexts & rename this to `MessagingProvider`.
Or, I wonder if we actually need this context? I know we wanted to handle most of the logic here, but since we couldn't do everything what we wanted with XMTP, I think this is a bit overkill now. We can just call these 3 hooks where we need them instead ;)

src/pages/_app.tsx
```diff
@@ -99,23 +100,25 @@ const App = ({
...
         >
           <SWRConfig value={{ fetcher: fetcherForSWR }}>
-            <WagmiConfig config={config}>
-              <PostHogProvider>
-                <IntercomProvider>
-                  <ExplorerProvider>
-                    <AppErrorBoundary>
-                      <Component {...pageProps} />
-                    </AppErrorBoundary>
+            <XMTPProvider>
```
BrickheadJohnny, line 103: Let's remove this provider and just use it inside the `MessagesSection` component & `SendNewMessage` modal. That way we won't load it by default, only when we actually need it.

src/pages/_error.tsx
```diff
@@ -84,13 +84,18 @@ const Page = ({ statusCode }): JSX.Element => {
   )
 }
 
-const PageWrapper = (props) => (
-  <IntercomProvider>
-    <Page {...props} />
-  </IntercomProvider>
-)
+const PageWrapper = (props) => {
+  console.error("PageWrapper _error.tsx")
```
BrickheadJohnny, line 88: Make sure to remove unnecessary logs when you finished debugging ;)

src/components/common/Layout/components/Account/components/Notifications/components/web3Inbox.ts
```diff
@@ -0,0 +1,104 @@
...
+    isWeb3InboxLoading: isLoading || isPreparing || isW3iAccountLoading,
+    web3InboxError: w3IAccountError ?? error,
+    isRegisteringWeb3Inbox,
+    isSigningWeb3Inbox,
+    isSubscribingWeb3Inbox,
+    subscribeWeb3Inbox,
+  }
+}
+
+export const useGetWeb3InboxMessages = () => {
```
BrickheadJohnny, line 90: I think we don't really need a custom hook for this, let's just use the `useNotifications` hook instead, since we only need that in one place.

src/components/common/Layout/components/Account/components/Notifications/components/web3Inbox.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const toast = useToast()
+  const showErrorToast = useShowErrorToast()
+  const { signMessageAsync } = useSignMessage()
+
+  const { prepareRegistration, isLoading: isPreparing } = usePrepareRegistration()
+  const { register, isLoading: isRegisteringWeb3Inbox } = useRegister()
+  const { address, isConnecting, isReconnecting, status } = useAccount()
+  const {
+    data: w3IAccount,
```
BrickheadJohnny, line 36: I think it's safe to call this `account`, we're inside the `useWeb3InboxMessaging` hook, so we know that this is the user's Web3Inbox account

src/components/common/Layout/components/Account/components/Notifications/components/xmtp.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export const useSubscribeXmtp = () => {
+  const [isSubscribingXmtp, setIsSubscribingXmtp] = useState<boolean>(false)
+
+  const toast = useToast()
+  const showErrorToast = useShowErrorToast()
+
+  const { data: signer, isLoading } = useWalletClient()
+
+  const subscribeXmtp = useCallback(async () => {
```
BrickheadJohnny, line 114: I wonder if we could use our `useSubmit` hook here? 👀 That would handle the loading state for us.

Another note: let's call it `subscribeToXMTP`, it feels more natural hehe

src/components/common/Layout/components/Account/components/Notifications/components/xmtp.ts
```diff
@@ -0,0 +1,142 @@
...
+  const { id } = useUser()
+
+  return async () =>
+    fetcherWithSign([
+      id ? `/v2/users/${id}/keys/1` : undefined,
+      { method: "DELETE" },
+    ])
+}
+
+export const useGetXmtpKeys = () => {
```
BrickheadJohnny, line 43: The hook returns the data/error & the loading state, so let's call it `useUsersXMTPKeys`, that's a more logical name in my opinion

src/components/common/Layout/components/Account/components/Notifications/components/xmtp.ts
```diff
@@ -0,0 +1,142 @@
...
+import { useFetcherWithSign } from "utils/fetcher"
+import { useAccount, useWalletClient } from "wagmi"
+
+export const useSaveXmtpKeys = () => {
+  const fetcherWithSign = useFetcherWithSign()
+  const { data: signer } = useWalletClient()
+
+  const { id } = useUser()
+
+  return useCallback(
```
BrickheadJohnny, line 15: I think we could use the custom `useSubmit` hook here, it would handle the loading/error state automatically ;)

src/components/common/Layout/components/Account/components/Notifications/MessageSubscriptionModal.tsx
```diff
@@ -0,0 +1,100 @@
+import {
```
BrickheadJohnny, line 1: I think we don't use this component, right? Let's delete it if we don't need it!

src/components/[guild]/messages/SendNewMessage/SendNewMessage.tsx
```diff
@@ -53,128 +97,154 @@ const SendNewMessage = (props: ButtonProps) => {
...
-        onClick={onOpen}
-        {...(!isStuck && {
-          color: textColor,
-          colorScheme: buttonColorScheme,
-        })}
-        {...props}
-      >
-        New message
-      </Button>
+      <ModalOverlay />
```
Comment (BrickheadJohnny) [inline src/components/[guild]/messages/SendNewMessage/SendNewMessage.tsx:122]: This pattern is pretty strange for me, I think we can leave the button, modal & modal content in the same component and just import the component dynamically only for admins

BrickheadJohnny (changes requested):
Wrote a couple of comments for you, if you have any questions, don't hesitate to ping me on Discord!

Also, a note:
Please refactor the `xmtp.ts` & `web3inbox.ts` files & make a standalone file for each hook instead. That way we can easily search for the hooks in our codebase.

src/components/[guild]/messages/SendNewMessage/SendNewMessage.tsx
Comment (gdn992) [reply src/components/[guild]/messages/SendNewMessage/SendNewMessage.tsx:123]: I think this page is just loaded when the user is administrator. And all implementation is dynamic.

src/components/common/Layout/components/Account/components/Notifications/MessageSubscriptionModal.tsx
gdn992: true

src/components/common/Layout/components/Account/components/Notifications/components/xmtp.ts
gdn992: done, i put the saving functions into the submit hook

src/components/common/Layout/components/Account/components/Notifications/components/MessagingContext.tsx
gdn992: At first, it seemed like a good idea, but now I think it's overkill yes.

src/pages/_error.tsx
gdn992: ofc

src/components/common/Layout/components/Account/components/Notifications/components/xmtp.ts
gdn992: can I use the useSubmit hook here also?

gdn992: I used useSubmit for handle the loading state of this checking xmtp hook

BrickheadJohnny:
Checked the PR with the new changes, here are some additional comments:

- the `xmtp.ts` file is still there, I'd suggest you to make a `useUsersXMTPKeys.ts` & a `useSubscribeToXMTP.ts` file and put them close to the component/hook where we use them, e.g. `useUsersXMTPKeys.ts` can go in the same folder where we have the `useSendMessage.ts` hook, because we only use it there, and you can put the `useSubscribeToXMTP.ts` inside the `...Notifications/hooks/useSubscribeToXMTP.ts` folder
- the `useXmtpAccessChecking` is unnecessary in my opinon, because the `useCanMessage` hook already contains an `isLoading` & `error` state
- we can simplify the `useSendMessage` hook by using a simple `useSubmit` (that way we wouldn't need to parse the signed payload to get some properties from it & then just use `fetcherWithSign` where we actually need to sign the payload
- please use consistent hook names, I can see `...Xmtp` and `...XMTP` in the names, let's use the uppercase version everywhere
- I think we can simplify the `subscribeToXMTP` function like this (removed the unnecessary async-await, since `Client.create` returns a promise:

```ts
const subscribeToXMTP = () =>
  Client.create(signer, {
    persistConversations: false,
    env: "production",
  }).then(() => onSuccess())
```

- the `web3inbox.ts` filename is a bit misleading, it only contains one hook, so let's call it `useWeb3InboxSubscription.ts` instead & use a default export inside it
- I still see that you renamed a couple of properties in some hooks, e.g. `isRegistering` -> `isRegisteringWeb3Inbox`. I think it's also unnecessary, because we can easily understand that if `isRegistering` comes from the `useWeb3InboxSubscription` hook, then it relates to `Web3Inbox`. If we really need to rename it (because we have 2 or more variables with the same name), we can just do it where we actually use the hook 
- please think about variable names a bit more carefully, e.g. the `isLoading` prop which comes from the `useCanMessage` hook is renamed to `isCheckingAccess`. At first, I thought this variable comes from the `useAccess` hook, so let's call it `isCanMessageLoading` (again, only if we really need to rename it)

src/components/[guild]/messages/hooks/useUsersXMTPKeys.ts
```diff
@@ -0,0 +1,40 @@
...
+import { useFetcherWithSign } from "utils/fetcher"
+
+export const useUsersXMTPKeys = () => {
+  const fetcherWithSign = useFetcherWithSign()
+  const [data, setData] = useState(null)
+  const [isLoading, setIsLoading] = useState(true)
+  const [error, setError] = useState(null)
+  const { id, error: userError, isLoading: isUserLoading } = useUser()
+
+  useEffect(() => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/messages/hooks/useUsersXMTPKeys.ts:12]: I think we could just use SWR here, couldn't we? We wouldn't need to handle the isLoading, error, and data states in that case.

src/components/common/Layout/components/Account/components/Notifications/MessagingSection.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const {
+    isLoading: isCheckingXMTPAccess,
+    onSubmit: checkXMTPAccess,
+    response: hasXMTPAccess,
+  } = useSubmit(async () => canMessageStatic(address), {
+    onError: () => showErrorToast("Error happened during checking XMTP access"),
+  })
+
+  useEffect(() => {
+    checkXMTPAccess()
```
BrickheadJohnny, line 56: I think at this point, this could also be a useSWR call instead of useSubmit & useEffect, right?

src/components/common/Layout/components/Account/components/Notifications/components/useWeb3InboxSubscription.ts
```diff
@@ -0,0 +1,87 @@
...
+    } catch (subscribeError) {
+      showErrorToast("Couldn't subscribe to Guild messages")
+    }
+  }
+
+  return {
+    subscriptionStatus,
+    isLoading: isLoading || isPreparing || isAccountLoading,
+    error: accountError ?? error,
+    isRegistering: isRegistering,
```
BrickheadJohnny, line 80: You could use the shortened version of this, like so:

```ts
return {
    subscriptionStatus,
    isLoading: isLoading || isPreparing || isAccountLoading,
    error: accountError ?? error,
    isRegistering,
    isSigning,
    isSubscribing,
    subscribe: handleSubscribe
}
```

src/components/common/Layout/components/Account/components/Notifications/MessagingSection.tsx
```diff
@@ -0,0 +1,156 @@
...
+    subscriptionStatus,
+    subscribe,
+  } = useWeb3InboxSubscription()
+  const { canMessageStatic } = useCanMessage()
+  const showErrorToast = useShowErrorToast()
+
+  const { address } = useAccount()
+
+  const {
+    isLoading: isCheckingXMTPAccess,
```
BrickheadJohnny, line 48: I'd call these `isCanMessageLoading`, `onCanMessageSubmit` (but you don't need this one, check my next comment!), `canMessage`

BrickheadJohnny (commented):
Now this looks good overall, but left a couple of more comments! ;)

Also, 2 more things:
- There's still now spacing between the "Select protocol" & "Recipient roles" inputs in the send message form
- We should also display some kind of info text for the users before they send an XMTP message for the first time, just to let them know that their keys will be stored in our backend!

After you fixed these, please request a review from @dovalid too 😉

src/components/common/Layout/components/Account/components/Notifications/MessagingSection.tsx
gdn992: I think "isCheckingXMTPAccess" accurately describes its purpose.  I think "CanMessageLoading" doesn't specify whether it's for XMTP or W3I.

BrickheadJohnny:
Closing it for now, we'll focus on stability & other features for now.

PR #1073, opened 2024-02-19
Set MIRROR_COLLECT as the default req type for Mirror

marclllaks:
- Small UI change
- `MIRROR_COLLECT` is the default req type for Mirror

Please run it, it works locally but this is my first frontend pr 🙈


BrickheadJohnny (approved):
🚀

PR #1079, opened 2024-02-23
Feat: Check if a user has already claimed a POAP

gdn992:
Check if a user has already claimed a POAP


src/components/[guild]/claim-poap/components/ClaimPoap.tsx
```diff
@@ -47,6 +49,8 @@ const ClaimPoap = ({ rolePlatformId }: Props) => {
         <ClaimPoapButton
           rolePlatformId={rolePlatformId}
           isDisabled={!isButtonEnabled}
+          isClaimed={claimed}
```
Comment (BrickheadJohnny) [inline src/components/[guild]/claim-poap/components/ClaimPoap.tsx:52]: I think we could move this logic inside `ClaimPoapButton` instead of using props

src/platforms/Poap/ShowMintLinkButton.tsx
```diff
@@ -0,0 +1,92 @@
...
+    modalProps: { isOpen, onOpen, onClose },
+  } = useClaimText(rolePlatformId)
+  const httpsLink = response?.uniqueValue?.replace("http://", "https://")
+
+  return (
+    <>
+      <Button onClick={onOpen} {...rest}>
+        {children}
+      </Button>
+      <Modal isOpen={isOpen} onClose={onClose}>
```
BrickheadJohnny, line 44: Now this modal is duplicated, could you please refactor it so we use the same modal both in the claim flow and in the "show mint link" flow?

BrickheadJohnny (changes requested):
Great PR overall, but I left 2 short comments! 😉

src/components/[guild]/claim-poap/components/ClaimPoap.tsx
```diff
@@ -48,9 +48,8 @@ const ClaimPoap = ({ rolePlatformId }: Props) => {
       >
         <ClaimPoapButton
           rolePlatformId={rolePlatformId}
+          rolePlatform={rolePlatform}
```
Comment (BrickheadJohnny) [inline src/components/[guild]/claim-poap/components/ClaimPoap.tsx:51]: I think we should only provider `rolePlatform` here, because then we can use `rolePlatform.id` inside this component

src/components/[guild]/claim-poap/components/MintLinksModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+  ModalContent,
+  ModalHeader,
+  ModalOverlay,
+  Spinner,
+  Text,
+} from "@chakra-ui/react"
+import { ArrowSquareOut, CheckCircle } from "phosphor-react"
+import ErrorAlert from "../../../common/ErrorAlert"
+
+export const MintLinksModal = ({ isOpen, onClose, isLoading, error, response }) => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/claim-poap/components/MintLinksModal.tsx:20]: nit, but let's call it `MintLinkModal` (both the file and the component), because we only show 1 mint link in it

BrickheadJohnny (approved):

PR #1080, opened 2024-02-23
chore: upgrade dependencies

BrickheadJohnny:


BrickheadJohnny:
For some reason, we got "too many open files" errors on Vercel with the upgraded deps, so we won't merge this. Working on a solution, will open another PR in a couple of weeks!

PR #1082, opened 2024-02-26
fix: crm fix navigating back

FBalint:


src/pages/_app.tsx
```diff
@@ -44,8 +44,14 @@ const App = ({
   const [isRouteChangeInProgress, setIsRouteChangeInProgress] = useState(false)
   const { colorMode } = useColorMode()
 
+  let previousPathname = null
```
dovalid, line 47: Shouldn't we use useRef or useState here? I haven't tried if it works yet, but I think we should since we're in a component and they also mention it in a comment on Stackoverflow: https://stackoverflow.com/questions/62368109/update-router-query-without-firing-page-change-event-in-next-js#comment119198796_62368514

src/pages/[guild]/members.tsx
```diff
@@ -141,17 +141,21 @@ const GuildPage = (): JSX.Element => {
...
-     * but I'm not sure the performance cost would worth handling this edge case, so
-     * left it this way for now
-     */
-    window.history.pushState("", "", `?${queryString}`)
-  }, [queryString])
+    if (!router.isReady) return
+
+    const searchParams = new URLSearchParams(queryString)
+    const query = {}
+    searchParams.forEach((value, key) => {
```
Comment (dovalid) [inline src/pages/[guild]/members.tsx:148]: Why do we have to rebuild the query? If it's just because it's returned stringified, we can change that by removing the `.toString` in `buildQueryStringFromState` and renaming it to `buildQueryFromState`, then applying `.toString` just when passing it to `useMembers`

src/pages/[guild]/members.tsx
```diff
@@ -141,17 +141,21 @@ const GuildPage = (): JSX.Element => {
...
+      query[key] = query[key] || []
+      query[key].push(value)
+    })
+
+    router.replace({
+      pathname: router.pathname,
```
Comment (dovalid) [inline src/pages/[guild]/members.tsx:154]: Do we need this line? IIRC we can just pass replace what we actually wan't to replace, no?

Comment (FBalint) [reply src/pages/[guild]/members.tsx:154]: We could instead of all this, simply just do
```
  useEffect(() => {
    if (!router.isReady || !queryString) return

    const asPath = router.asPath.split("?")[0]
    router.replace(`${asPath}?${queryString}`)

  }, [queryString, router.isReady])
```

I did not want to depend on string splitting, and then it just escalated.

src/pages/[guild]/members.tsx
Comment (FBalint) [reply src/pages/[guild]/members.tsx:148]: It was because `router.replace` requires a query object, where keys are the names of the query parameters and values are the corresponding values. URLSearchParams is not compatible with this format.

Also, I didn't realise that we only use `buildQueryStringFromState` in this component. So maybe this query building logic could be moved there instead.

dovalid (approved):

PR #1083, opened 2024-02-26
fix: undefined platformType in EditRewardAvailabilityModal & enable adding POAPs in the edit/create role drawers

gdn992:


BrickheadJohnny (commented):
Looks good to me, just one tiny thing I noticed: if I don't upload mint links during the POAP creation flow and then I open the `EditRewardAvailabilityModal` from the Edit Role drawer, the capacity input will be in error state and I can't close the modal by clicking on the "Done" button because of that. Could you please check what's the issue with this? 👀

BrickheadJohnny (approved):

PR #1085, opened 2024-02-26
Fix: GuildSelect: current guild as defaultValue

gdn992:


src/requirements/common/GuildSelect.tsx
```diff
@@ -33,47 +40,32 @@ const GuildSelect = ({ baseFieldPath }) => {
 
   const { data: guildOptions, isValidating: isGuildsLoading } =
     useGuilds(debouncedSearchValue)
+  const { isLoading, ...currentGuild } = useGuild()
 
-  const guildId = useWatch({ name: `${baseFieldPath}.data.guildId` })
-
-  const { isLoading, ...fetchedGuild } = useGuild(
```
BrickheadJohnny, line 39: I think we need this logic so users can search for hidden guilds too ;)

BrickheadJohnny (changes requested):
Nice, clean solution, but I think you accidentally removed some code which we actually need, pls double check!

src/requirements/common/GuildSelect.tsx
gdn992: Uhh true, i put it back

BrickheadJohnny (approved):

PR #1091, opened 2024-02-28
fix: dont auto prepend guildxyz to links in secret rewards

FBalint:


TomiOhl (commented):
Can confirm this does the job for me 👍🏻 
(I'll leave the actual approval to a frontend colleague though)

BrickheadJohnny (commented):
The logic itself looks good, I wonder if we could use react-hook-form to handle the link editor input state? That way we could also use the `FormErrorMessage` component for the error message & we could also just write a validation rule + we wouldn't need to store this value in state

src/components/lexical/plugins/ToolbarPlugin/components/LinkEditor.tsx
```diff
@@ -99,70 +124,83 @@ const LinkEditor = ({ isOpen, onOpen, onClose, insertLink }: LinkEditorProps) =>
...
+            fontSize="sm"
+            border="none"
+          >
+            Edit link
+          </PopoverHeader>
+
+          <PopoverArrow />
+          <PopoverCloseButton />
+          <PopoverBody p={2}>
+            <form
```
BrickheadJohnny, line 167: You could use `methods.handleSubmit` on the button instead (the callback which you pass to this function will receive the form state as a param), in that case you wouldn't need to use `useWatch` ;)

src/components/lexical/plugins/ToolbarPlugin/components/LinkEditor.tsx
```diff
@@ -27,33 +31,54 @@ type LinkEditorProps = {
...
+  const link = useWatch({ name: "link", control: control })
+
   const [editor] = useLexicalComposerContext()
 
   const initialFocusRef = useRef<HTMLInputElement>()
-  const [linkUrl, setLinkUrl] = useState("")
   const [lastSelection, setLastSelection] = useState(null)
   const [shouldOpenEditor, setShouldOpenEditor] = useState(false)
 
+  const checkLinkValid = (linkToCheck: string) => {
```
BrickheadJohnny, line 52: Since this function don't have any dependencies, just a simple param, you could move it outside of the component

BrickheadJohnny (changes requested):
Left 2 more small comments ;)

BrickheadJohnny (approved):

PR #1095, opened 2024-02-28
feat: farcaster accept urls too

FBalint:


src/requirements/Farcaster/FarcasterRequirement.tsx
```diff
@@ -47,11 +47,24 @@ const FarcasterRequirement = (props: RequirementProps) => {
...
-                <DataBlockWithCopy text={requirement.data.hash}>
-                  {shortenHex(requirement.data.hash, 3)}
-                </DataBlockWithCopy>
-                {` cast`}
+                {requirement.type === "FARCASTER_LIKE" ? "Like" : "Recast"}
+
+                {requirement.data?.hash && (
+                  <>
+                    {` the `}
+                    <DataBlockWithCopy text={requirement.data.hash}>
```
BrickheadJohnny, line 55: Let's use a link here + I think in case the user provided a hash, we could fetch the cast info with the `https://api.neynar.com/v2/farcaster/cast/?api_key=...&viewer_fid=-1&type=hash&identifier=...` endpoint, and display a Warpcast link in that case too ;)

BrickheadJohnny (changes requested):

src/requirements/Farcaster/FarcasterRequirement.tsx
```diff
@@ -3,14 +3,19 @@ import Requirement, {
...
+import FarcasterCast from "./components/FarcasterCast"
+import useFarcasterCast from "./hooks/useFarcasterCast"
 import { useFarcasterUser } from "./hooks/useFarcasterUsers"
 
 const FarcasterRequirement = (props: RequirementProps) => {
   const requirement = useRequirementContext()
 
   const { data: farcasterUser } = useFarcasterUser(requirement.data?.id)
+  const {
+    data: farcasterCast,
```
BrickheadJohnny, line 15: nit: I'd call these cast, isCastLoading & castError, just for consistency

src/requirements/Farcaster/hooks/useFarcasterCast.ts
```diff
@@ -0,0 +1,37 @@
...
+import useSWRImmutable from "swr/immutable"
+
+const BASE_URL =
+  "https://api.neynar.com/v2/farcaster/cast?api_key=NEYNAR_API_DOCS&identifier="
+
+const useFarcasterCast = (hash?: string, url?: string) => {
+  const debouncedHash = useDebouncedState(hash)
+  const debouncedUrl = useDebouncedState(url)
+
+  const fetchUrl = debouncedHash
```
BrickheadJohnny, line 11: Consider validating the hash & url here, becuase I managed to run into this issue, hehe:
image
(we shouldn't fetch anything if we pass invalid data)

BrickheadJohnny (commented):
OMG, this is awesome! 🔥 
Pushed one refactor & left 2 minor comments, please check them out, but fantastic work overall! 💯

src/requirements/Farcaster/hooks/useFarcasterCast.ts
```diff
@@ -0,0 +1,37 @@
+import useDebouncedState from "hooks/useDebouncedState"
+import useSWRImmutable from "swr/immutable"
+
+const BASE_URL =
+  "https://api.neynar.com/v2/farcaster/cast?api_key=NEYNAR_API_DOCS&identifier="
+
+const useFarcasterCast = (hash?: string, url?: string) => {
+  const debouncedHash = useDebouncedState(hash)
```
BrickheadJohnny, line 8: Oh, one more comment, I think we should use the `useDebouncedState` hook inside the form component, so we don't use the debounced values in the requirement display component ;)

src/requirements/Farcaster/FarcasterRequirement.tsx
```diff
@@ -24,6 +25,14 @@ const FarcasterRequirement = (props: RequirementProps) => {
       : undefined
   )
```
dovalid, line 27: Since we need different hooks for the different types of the req, we should create separate components for these cases, like in PointsRequirement or GuildRequirement. That way we wouldn't need to have all hooks mounted in all cases and the conditions duplicated for them so they don't actually fetch if they don't have to

(We should have done this when we've added channels actually, but it's even more relevant now that we're adding casts too)

src/requirements/Farcaster/components/FarcasterCastHash.tsx
```diff
@@ -14,28 +17,65 @@ const FarcasterCastHash = ({ baseFieldPath }: Props) => {
...
+            },
+          })}
+        />
+
+        <FormErrorMessage>
+          {!hash && parseFromObject(errors, baseFieldPath)?.data?.url?.message}
+        </FormErrorMessage>
+      </FormControl>
+
+      <FormControl
```
dovalid, line 55: Do we still need the hash input? If so, ideally we should merge the logics into one input so it works with both as `Cast URL or hash`. Now I find it confusing that at first it looks like I have to provide both. And we should display the FarcasterCast below that when filled, as it depends on it's value (it looks amazing btw!)

src/requirements/Farcaster/components/FarcasterCast.tsx
```diff
@@ -0,0 +1,155 @@
...
+        <Text fontSize="xs" opacity={0.6}>
+          {prettyDate}
+        </Text>
+      </Stack>
+      <HStack spacing={3} ml="auto">
+        <Stat icon={Heart} value={cast.likes} />
+        <Stat icon={ShareNetwork} value={cast.recasts} />
+        <Stat icon={Chat} value={cast.replies} />
+      </HStack>
+      <Icon as={CaretRight} />
```
dovalid, line 115: We should use ArrowSquareOut here imo, that expresses better that it's an external link and we follow that pattern in the app (also CaretRight for same-page navigation, and ArrowSquareIn for modals - this latter one is not ideal tho)

dovalid (changes requested):

src/requirements/Farcaster/components/FarcasterCastHash.tsx
dovalid: Oh I see that we also save them differently and it'd be complicated to reverse fetch what to display. Can't we just always save the hash (I think that's safer than an url?) and use the url to get the hash and vica versa?

BrickheadJohnny: I ended up implementing another solution, because supporting both hashes and urls seamlessly with 1 input required a pretty hacky solution. So, I added an uncontrolled input for the URL: if the user pastes a valid URL & we can fetch the hash, we'll set `data.hash` to the fetched one, otherwise we just display an error.

PR #1099, opened 2024-03-01
feat: gather town space as reward

FBalint:
TODO:

- [x] create and edit form button disable state control
- [x] api key field should be optional on edit (also depends on ongoing backend discussions)
- [x] refactoring


src/platforms/Gather/GatherConnectionStatusAlert.tsx
```diff
@@ -0,0 +1,89 @@
...
+          </>
+        ),
+      }
+    } else {
+      return {
+        status: "info",
+        children: (
+          <>
+            <AlertIcon mt={0} />
+            <p>Enter API key and space URL below to set up connection.</p>
```
TomiOhl, line 67: Nit
```suggestion
            <p>Enter your API key and space URL below to set up a connection.</p>
```

src/platforms/Gather/GatherForm.tsx
```diff
@@ -0,0 +1,20 @@
+import { Divider, Text } from "@chakra-ui/react"
+import GatherConnectionForm from "./GatherConnectionForm"
+import GatherGuestForm from "./GatherGuestForm"
+
+const GatherForm = () => {
+  return (
+    <>
+      <Text colorScheme="gray" fontWeight="semibold" mb="8">
+        The email addresses of eligible users will be added to the guest list for the
+        Gather space, which you can specify below.
```
TomiOhl, line 10:
```suggestion
        The email addresses of eligible users will be added to the guest list of the
        Gather space, which you can specify below.
```

src/platforms/Gather/GatherConnectionForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+      <GatherConnectionStatusAlert
+        isLoading={accessLoading}
+        success={accessSuccess}
+        error={accessError}
+      />
+
+      <FormControl pt={{ md: 0.5 }} isInvalid={!!errors?.gatherApiKey}>
+        <Stack>
+          <HStack>
+            <Text as="span">API key:</Text>
```
dovalid, line 52: Should use FormLabels, so the styling is consistent and the input is focusable by clicking on the label too

src/platforms/Gather/GatherConnectionForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+          <Link
+            href="https://gather.town/apiKeys"
+            textDecoration="underline"
+            isExternal
+          >
+            here
+          </Link>
+          .
+        </Text>
+        <FormErrorMessage>{errors?.gatherApiKey?.message}</FormErrorMessage>
```
dovalid, line 85: Should use the common FormErrorMessage component, so it animates

src/platforms/Gather/GatherConnectionForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+              {...register("gatherApiKey", {
+                required: "This field is required",
+              })}
+            />
+            <InputRightElement onClick={() => setShowApiKey(!showApiKey)}>
+              {showApiKey ? <EyeClosed /> : <Eye />}
+            </InputRightElement>
+          </InputGroup>
+        </Stack>
+        <Text as="small" colorScheme="gray">
```
dovalid, line 74: Should use FormHelperText

src/platforms/Gather/GatherConnectionForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+          <InputGroup>
+            <Input
+              autoComplete="off"
+              type={showApiKey ? "text" : "password"}
+              {...register("gatherApiKey", {
+                required: "This field is required",
+              })}
+            />
+            <InputRightElement onClick={() => setShowApiKey(!showApiKey)}>
+              {showApiKey ? <EyeClosed /> : <Eye />}
```
dovalid, line 70: Would be great to wrap the icons in an IconButton so we have hover/focus states and they're focusable. Or at least set a cursor: pointer style

src/platforms/Gather/GatherConnectionForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+              placement="top"
+              hasArrow
+            >
+              <Icon as={Question} color="GrayText" />
+            </Tooltip>
+          </HStack>
+          <InputGroup>
+            <Input
+              autoComplete="off"
+              type={showApiKey ? "text" : "password"}
```
dovalid, line 64: When being a pw and clicking next, the browser asked me if I'd want to save it, idk if we can prevent that

src/platforms/components/GatherPreview.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { useWatch } from "react-hook-form"
+import PlatformPreview from "./PlatformPreview"
+
+const GatherPreview = () => {
+  const platformGuildData = useWatch({
+    name: "rolePlatforms.0.guildPlatform.platformGuildData",
+  })
+
+  return (
+    <PlatformPreview
```
dovalid, line 11: AvailabilitySetup is not shown for Gather, while it should be. We have to add it elsewhere, just commenting here

src/platforms/Gather/GatherConnectionForm.tsx
FBalint: The `autoComplete='off'` attribute should theoretically address this issue, yet it appears to be ineffective. Should we then default to a standard text input?

Most solutions to this often seem excessively complicated. We could apply the `-webkit-text-security: disc` property in CSS, which appears to be widely supported on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-security#browser_compatibility) and [Can I Use](https://caniuse.com/?search=-webkit-text-security), though it's still considered non-standard.

src/platforms/Gather/GatherGuestForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { SectionTitle } from "components/common/Section"
+import { Question } from "phosphor-react"
+import { useFormContext } from "react-hook-form"
+
+const GatherGuestForm = () => {
+  const { register } = useFormContext<AddGatherFormType>()
+
+  return (
+    <>
+      <FormControl>
```
dovalid, line 20: We shouldn't use FormControl as a high level wrapper like this and have multiple inputs in one. We should replace the Stacks around the labels + inputs below instead to be individual FormControls

Issue #1102, opened 2024-03-05
Can not setup in local machine


oneforalone:
Some changes regarding the SDK installation workflow:
- The `WAAS_WEB_URL` and `WAAS_VIEM_URL` env vars have to be defined in `.env.local`
   - Value is available on 1Password, and from Vercel
- If the env vars are defined, it is enough to run `npm i`, which will trigger the `postinstall` script, which installs the SDK-s if the env vars are defined correctly
   - Everyone within the org will get the env vars
- For external developers, who don't have the env vars, running `git apply no_waas.diff` should make the appropriate changes, so the app can be built without the SDKs
   - We could also just run a `git revert` for the PR's squashed commit
   - These git-based solutions might run into conflicts later on, but hopefully the SDKs get published on NPM soon enough, so it might not cause problems

_Originally posted by @cs-balazs in https://github.com/guildxyz/guild.xyz/issues/1015#issuecomment-1908302579_

oneforalone:
I can not make it on my macbook m2 cause `git apply no_waas.diff` no working, it complains:
```
error: patch failed: src/components/[guild]/JoinModal/hooks/useConnectPlatform.ts:1
error: src/components/[guild]/JoinModal/hooks/useConnectPlatform.ts: patch does not apply
error: patch failed: src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx:32
error: src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx: patch does not apply
error: patch failed: src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/GoogleLoginButton/GoogleLoginButton.tsx:1
error: src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/GoogleLoginButton/GoogleLoginButton.tsx: patch does not apply
error: patch failed: src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/GoogleLoginButton/components/CopyCWaaSBackupData.tsx:1
error: src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/GoogleLoginButton/components/CopyCWaaSBackupData.tsx: patch does not apply
error: patch failed: src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/GoogleLoginButton/components/UserOnboardingModal.tsx:1
[...]

PR #1104, opened 2024-03-05
WaaS mock

cs-balazs:
Original report: https://github.com/guildxyz/guild.xyz/issues/1102

The `no_waas.diff` method is already broken, because the diff conflicts
This PR adds a `mockWaas.js`, which instead of removing the waas-related code, writes mocks in place of the two relevant packages

Thanks for the report @oneforalone! Can you please check if this method works for you?


BrickheadJohnny (approved):

oneforalone:
Awesome, it works now. Thanks for your quick response.

PR #1111, opened 2024-03-08
fix: users should be able to copy text from select inputs

FBalint:


src/components/common/StyledSelect/components/CustomIndicatorsContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+}: {
+  value: StyledSelectProps["value"]
+  pathToCopy: string
+} & IndicatorsContainerProps<unknown, boolean, GroupBase<unknown>>) => {
+  const [copied, setCopied] = useState(false)
+  const successIconColor = useColorModeValue("green.800", "green.200")
+
+  const handleCopyToClipboard = async () => {
+    try {
+      await navigator.clipboard.writeText(parseFromObject(value, pathToCopy))
```
BrickheadJohnny, line 26: Is it intentional that you haven't used the `useClipboard` hook here?

src/components/common/StyledSelect/components/CustomIndicatorsContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+      }, 1000)
+    } catch (err) {
+      console.error("Failed to copy:", err)
+    }
+  }
+
+  return (
+    <components.IndicatorsContainer {...rest}>
+      {value && (
+        <MotionIconButton
```
BrickheadJohnny, line 39: Do we really need a motion component here? 👀

src/components/common/StyledSelect/components/CustomIndicatorsContainer.tsx
FBalint: Wow I missed this, thanks!

src/components/common/StyledSelect/components/CustomIndicatorsContainer.tsx
```diff
@@ -0,0 +1,43 @@
...
+          backgroundColor={
+            hasCopied && transparentize(successIconColor, 0.12)(theme)
+          }
+          color={hasCopied && successIconColor}
+          icon={hasCopied ? <Check /> : <Copy />}
+          aria-label={"Copy"}
+          size={"xs"}
+          rounded={"full"}
+          variant={"ghost"}
+          onMouseDown={() => {
```
BrickheadJohnny, line 33: maybe `onClick` + `e.stopPropagation()` would work properly and then we wouldn't need to use `onMouseDown` here ;) Please check and lmk!

FBalint: I found that react-select uses `onMouseDown` to set the dropdown state, which happens before `onClick`. So we kind of need to get ahead of it by using onMouseDown here as well

BrickheadJohnny: Hmm, I see, thanks for the explanation.
The problem with the current solution is that it won't work on mobile (or at least I couldn't really click on the copy button), so let's brainstorm about this issue on Discord ;)

BrickheadJohnny (approved):

PR #1114, opened 2024-03-11
Feat: form related activity log actions

gdn992:


src/components/[guild]/activity/ActivityLogAction/components/FormTag.tsx
```diff
@@ -0,0 +1,136 @@
...
+type Props = FormTagProps & Omit<TagProps, "colorScheme">
+
+type FormTagProps = { formId: number; guildId: number }
+
+type ClickableFormTagProps = FormTagProps & {
+  activityLogType?: ActivityLogType
+  userId?: number
+}
+
+type FormField = {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/activity/ActivityLogAction/components/FormTag.tsx:31]: We have types for these in the `@guildxyz/types` package, so let's use those! (`Schemas["Form"]` & `Schemas["Field"]`)

src/components/[guild]/activity/ActivityLogAction/components/FormTag.tsx
```diff
@@ -0,0 +1,136 @@
...
+            <Circle bgColor={imgBgColor} size={4}>
+              <PencilSimpleLine />
+            </Circle>
+            <Text as="span" w="max-content">
+              {data?.name}
+            </Text>
+          </HStack>
+        )}
+
+        <TagRightIcon as={DotsThreeVertical} />
```
Comment (BrickheadJohnny) [inline src/components/[guild]/activity/ActivityLogAction/components/FormTag.tsx:80]: We should display the icon conditionally, just like we did inside `GuildTag` (we don't need to render it when the tag is in the filters bar)

image

src/components/[guild]/activity/ActivityLogAction/components/FormTag.tsx
```diff
@@ -0,0 +1,136 @@
...
+} from "../../ActivityLogFiltersBar/components/ActivityLogFiltersContext"
+import ClickableTagPopover from "./ClickableTagPopover"
+import ViewInFormResponses from "./ClickableTagPopover/components/ViewInFormResponses"
+
+type Props = FormTagProps & Omit<TagProps, "colorScheme">
+
+type FormTagProps = { formId: number; guildId: number }
+
+type ClickableFormTagProps = FormTagProps & {
+  activityLogType?: ActivityLogType
```
Comment (BrickheadJohnny) [inline src/components/[guild]/activity/ActivityLogAction/components/FormTag.tsx:27]: I think this data is available in context, so let's use it from there instead of passing down here

BrickheadJohnny (changes requested):

BrickheadJohnny (approved):

PR #1118, opened 2024-03-12
refactor: call api when pressing reset to original button

FBalint:


src/components/[guild]/Requirements/components/ResetRequirementButton.tsx
```diff
@@ -1,25 +1,29 @@
...
+      size={"sm"}
+      isDisabled={isLoading}
+      onClick={() =>
+        onSubmit({
+          ...requirement,
+          data: { ...requirement.data, customName: "", customImage: "" },
+        })
+      }
+    >
+      {isLoading ? (
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/components/ResetRequirementButton.tsx:20]: We could just use the `Button` component's `isLoading` prop, that way we'd have a spinner by default & we wouldn't need to manually handle the `isDisabled` state ;)

BrickheadJohnny (changes requested):
Very nice cleanup, noticed one small issue, but otherwise LGTM!

src/requirements/VisitLink/VisitLinkRequirement.tsx
```diff
@@ -97,7 +100,7 @@ const VisitLinkRequirement = ({ ...props }: RequirementProps) => {
                 <Text wordBreak="break-word" flexGrow={1}>
                   <Original />
                 </Text>
-                {!!formContext && <ResetRequirementButton />}
+                <ResetRequirementButton />
```
dovalid, line 103: The condition was there so we only show the reset button in the edit drawer, so in normal view the admin sees the same as every other member. Is there a particular reason you've removed it, or was it just unclear to you? (In that case, we should add a comment about it)

dovalid: Other than this, approved

BrickheadJohnny (approved):
Made some minor changes in the `RequirementNameEditor` component, lmk if you guys like it or not. The PR itself is approved by me too!

src/requirements/VisitLink/VisitLinkRequirement.tsx
FBalint: Thanks, I believe it was accidental, whilst trying to figure out what it was doing. I added a comment to make it more clear!

PR #1127, opened 2024-03-18
fix: scroll restoration in guild explorer

FBalint:
I implemented a custom scroll restoration for the explorer page.

The built-in scroll restoration functioned only with backward navigation, such as using `router.back()` or the browser's back button. However, our "Back to explorer" link doesn't operate in this manner; it directs us to the `/explorer` URL instead of navigating back.

Additionally, the standard scroll restoration is affected by the "Your guilds" list, as it loads after the scroll position has been restored, altering the expected position.

Note that I had to implement an undesirable delay in the manual scroll restoration. This is because the built-in restoration and our custom solution currently execute simultaneously. Without this delay, the built-in mechanism would override our manual approach.

If the built-in scroll restoration is not essential elsewhere, I recommend disabling it. This change would allow us to eliminate the delay

### DIFF
--- a/src/components/common/Layout/components/NavMenu/NavMenu.tsx
+++ b/src/components/common/Layout/components/NavMenu/NavMenu.tsx
@@ -12,7 +12,10 @@ import {
   useColorModeValue,
 } from "@chakra-ui/react"
 import Button from "components/common/Button"
+import { useSetAtom } from "jotai"
 import dynamic from "next/dynamic"
+import { useRouter } from "next/router"
+import { ExplorerScrollRestoration } from "pages/explorer"
 import {
   CircleWavyCheck,
   Code,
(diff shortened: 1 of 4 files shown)


dovalid (commented):
Nice, it works much better now! Also it's a nice addition that it works on the Back to explorer button too.

For the Your guilds section, can't we achieve that it wouldn't only load on mount and cause a layout shift, but be already there when going back?

I wouldn't disable the auto scroll-restoration, it could be useful in other places too, like navigating to a form / NFT page from a guild page then going back, or to CRM from activity log with View in members button, etc. It's another question that it often jumps in these cases too because of the client side rendering.. 🤔

For the current solution, my only suggestion is to extract the related logic from explorer to a useScrollRestoration() hook

BrickheadJohnny (approved):

PR #1129, opened 2024-03-19
Twitter v1 cleanup + enable

cs-balazs:
- Enables TWITTER_V1 `isPlatform`
- Removes `TwitterV1Tooltip`
- Disables Twitter `isPlatform`
- Removes now unnecessary API endpoint, which generated Twitter v1 OAuth tokens


BrickheadJohnny (approved):
Lovely PR, I just wonder if we need to filter out rewards with `isPlatform:false` from the `AccountConnections` component too? Because right now if someone already connected with Twitter V2, they'll see both that connection (since that data comes from our API) and the new (v1) connect button in the Account modal.

cs-balazs:
@BrickheadJohnny Nice idea! Changed in https://github.com/guildxyz/guild.xyz/pull/1129/commits/464710dcaf696003078d2ed39e522b74d50064e1!

PR #1131, opened 2024-03-19
fix(access check): check for falsy value instead of `null`

BrickheadJohnny:
Gate will return `access:false` with `PLATFORM_CONNECT_INVALID` & `PLATFORM_NOT_CONNECTED` errors from now, so we should check for falsy values (instead of `null`) where we use requirement access in if statements / filters / etc.


cs-balazs (approved):
🚀

PR #1134, opened 2024-03-22
refactor: remove deprecated `/guilds` mutations & improve the existing ones

BrickheadJohnny:


src/components/[guild]/EditGuild/hooks/useEditGuild.ts
```diff
@@ -393,4 +402,16 @@ const useEditGuild = ({ onSuccess, guildId }: Props = {}) => {
...
 
+const mutateGuildsCache = (prev: GuildBase[], editedGuild: GuildBase) =>
+  prev?.map((_guild) => {
+    if (_guild.id !== editedGuild.id) return _guild
+    return {
+      ..._guild,
+      name: editedGuild.name,
+      imageUrl: editedGuild.imageUrl,
+      urlName: editedGuild.urlName,
+      hideFromExplorer: editedGuild.hideFromExplorer,
```
Comment (cs-balazs) [inline src/components/[guild]/EditGuild/hooks/useEditGuild.ts:413]: Would it cause any issues if we spread `editedGuild` as well? We might forget to add here any new fields, we might add later

Comment (BrickheadJohnny) [reply src/components/[guild]/EditGuild/hooks/useEditGuild.ts]: Hmm, I think we don't actually get back a `GuildBase` when editing a guild, hehe, that's why I implemented it this way, but actually, that shouldn't be a problem, I agree with you! Fixed in 2172c2243e4a2d1771f79d69546e5091101ae8c7!
We'll need to fix these types at some points... 😂

cs-balazs (approved):
I haven't really tested it, but the changes look great to me! 🚀

PR #1135, opened 2024-03-25
[GUILD-2561] - Email allowlist

cs-balazs:
- `RequirementAccessIndicator`: Turned out that a missing email is also returned as a platform error, so I added a condition for that as well. Started to get ugly, so added a const for the messages
- Added a `type` selector for the allowlist form with options `EVM and Fuel addresses` and `Email addresses`
- Validation uses `z.string().email()` just to be sure that it validates using the same email rules, as the schema used on BE
- Companion PRs:
  - Gate: https://github.com/guildxyz/guild-gate/pull/715
  - Core: https://github.com/guildxyz/guild-backend/pull/1369

### DIFF
--- a/src/requirements/Allowlist/AllowlistForm.tsx
+++ b/src/requirements/Allowlist/AllowlistForm.tsx
@@ -82,9 +82,7 @@ const AllowlistForm = ({
         ?.toString()
         ?.split("\n")
         ?.filter((line) => !!line)
-        ?.map((line) =>
-          type === "ALLOWLIST_EMAIL" ? line.trim() : line.slice(0, 42)
-        )
+        ?.map((line) => line.trim())
 
       setValue(`${baseFieldPath}.data.addresses`, lines, { shouldValidate: true })
     }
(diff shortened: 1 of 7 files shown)


dovalid (approved):

src/requirements/Allowlist/AllowlistForm.tsx
```diff
@@ -43,16 +82,40 @@ const AllowlistForm = ({ baseFieldPath }: RequirementFormProps): JSX.Element =>
         ?.toString()
         ?.split("\n")
         ?.filter((line) => !!line)
-        ?.map((line) => line.slice(0, 42))
+        ?.map((line) =>
+          type === "ALLOWLIST_EMAIL" ? line.trim() : line.slice(0, 42)
```
BrickheadJohnny, line 86: I think this will "break" Fuel addresses (those are 66 long incl. 0x afaik) :ohno:

cs-balazs: I'm not sure if it currently works on prod then :D Should we just .trim always?

BrickheadJohnny: Oh 😂  Yeah, then let's just trim them 👍

cs-balazs: Made the change in https://github.com/guildxyz/guild.xyz/pull/1135/commits/7543acfad1b80896fb5735a0d7b6d9a7851031b2 👍

PR #1138, opened 2024-03-26
feat: use the `/role/:roleId/requirements` endpoint

BrickheadJohnny:
**Feature**
- added a new `useRequirements` hook
- using this new hook everywhere (where we used requirements from the `/guild-page` response before)
- refactored some mutations: we only need to mutate the requirements SWR cache after requirement cruds
- simplified the `CompleteCaptchaJoinStep` & `ConnectPolygonIDJoinStep` components (note: we shouldn't trigger a membership update after connecting PolygonID, but I'll leave that issue for a future PR, I'd like if we could store PolygonID connections in 1 place anyways...)

**Misc**
- fixed `LogicFormControl`, it didn't work properly in 1 case since we updated the requirement edit flow
- replaced `useNonPurchasableAssets` with a simple SWR hook


src/components/[guild]/Requirements/RoleRequirements.tsx
```diff
@@ -40,12 +41,13 @@ const RoleRequirements = ({
...
 }: Props) => {
   const guild = useGuild()
+  const { data } = useRequirements(role.id)
 
   const requirements =
     role.hiddenRequirements ||
-    ((role.requirements ?? []).length === 0 && !(guild as any).isFallback)
-      ? [...role.requirements, { type: "HIDDEN", roleId: role.id } as Requirement]
-      : role.requirements
+    ((data ?? []).length === 0 && !(guild as any).isFallback)
```
Comment (cs-balazs) [inline src/components/[guild]/Requirements/RoleRequirements.tsx:48]: Nit: Noticing a lot of `data?... ?? []` and `data ?? []` patterns, maybe it is worth considering for `useRequirements` to always return an array

src/components/create-guild/Requirements/EditRequirements.tsx
```diff
@@ -1,29 +1,33 @@
...
-type Props = { requirements: Requirement[] }
+type Props = { roleId: number }
 
-const EditRequirements = ({ requirements }: Props) => {
+const EditRequirements = ({ roleId }: Props) => {
+  const { data: requirements } = useRequirements(roleId)
   const logic = useWatch({ name: "logic" })
   const freeEntry = requirements.some(({ type }) => type === "FREE")
 
+  const mappedRequirements = requirements.map(mapRequirement)
```
cs-balazs, line 22: No optional chaining seems a bit scary here 👀

cs-balazs (commented):
Nice changes! Like how many mutations got a lot more readable! Left two comments, but they point out the same thing

src/components/create-guild/Requirements/EditRequirements.tsx
BrickheadJohnny: Oh, good point, added a `?`, thanks!

src/components/[guild]/Requirements/RoleRequirements.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/Requirements/RoleRequirements.tsx]: Hmm, not sure about this, I think in general data should be `undefined` until we don't fetch anything (or if we encounter an error), so I wouldn't like to assign a default value to it, but I assigned a default value to `data` in this component only, so the code is a bit cleaner now ;)

src/components/create-guild/Requirements/components/ExistingRequirementEditableCard.tsx
```diff
@@ -54,9 +54,8 @@ const ExistingRequirementEditableCard = ({
   const showViewOriginal =
     requirement.data?.customName || requirement.data?.customImage
 
-  const { roles } = useGuild()
-  const role = roles.find((r) => r.id === requirement.roleId)
-  const isLastRequirement = role.requirements.length === 1
+  const { data: requirements } = useRequirements(requirement.roleId)
+  const isLastRequirement = requirements.length === 1
```
cs-balazs, line 58: We might need a `?` here as well :ayy:

cs-balazs (approved):
Amazing changes 🚀

PR #1141, opened 2024-03-28
fix(join flow): join guild with hidden roles

BrickheadJohnny:
How to reproduce: 
- create a guild with a hidden free entry role
- join the guild with a user (who is not an admin in the guild)


cs-balazs (commented):
As I understand, the reason is that `roleAccesses` doesn't contain the hidden one? In that case, it lgtm 🚀

BrickheadJohnny:
> As I understand, the reason is that `roleAccesses` doesn't contain the hidden one? In that case, it lgtm 🚀

Yup, exactly!

cs-balazs (approved):

PR #1145, opened 2024-04-04
Google login force connection & mutation fix

cs-balazs:
- Perform forced (`disconnectFromExistingUser`) connection
  - This way, the Google account will always be linked to the user with the WaaS wallet
- Fixes a problem, when the Google account didn't immediately appear in the account modal after an initial login
  - This bug was likely introduced in the OAuth rework PR, probably the hook performed the mutation before that rework. Now we perform the mutation after the connection in this `useSubmit` with the `userId` taken from the keypair submission request

### DIFF
--- a/src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/GoogleLoginButton/hooks/useLoginWithGoogle.ts
+++ b/src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/GoogleLoginButton/hooks/useLoginWithGoogle.ts
@@ -7,6 +7,7 @@ import useSubmit from "hooks/useSubmit"
 import useToast from "hooks/useToast"
 import { useAtomValue, useSetAtom } from "jotai"
 import { useState } from "react"
+import { mutate } from "swr"
 import { shouldUseReCAPTCHAAtom } from "utils/recaptcha"
 import { useConnect } from "wagmi"
 import {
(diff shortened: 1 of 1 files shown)


src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/GoogleLoginButton/hooks/useLoginWithGoogle.ts
```diff
@@ -152,24 +153,43 @@ const useLoginWithGoogle = () => {
...
           ...signProps,
         },
         platformName: "GOOGLE",
         authData,
+        disconnectFromExistingUser: true,
       })
 
+      if (userId && platformUser) {
+        await mutate(
+          [`/v2/users/${userId}/profile`, { method: "GET", body: {} }],
```
BrickheadJohnny, line 184: Totally irrelevant for this PR, but I wonder if we should create a function which generates this SWR key? I noticed that we use it in multiple places, maybe it would make sense to store it in one place, so in case we change it in the future, we won't break anything.

BrickheadJohnny (approved):
🚀

src/components/_app/Web3ConnectionManager/components/WalletSelectorModal/components/GoogleLoginButton/hooks/useLoginWithGoogle.ts
cs-balazs: Hmm, yeah, that might be worth implementing later on! 💯

PR #1147, opened 2024-04-05
feat: access check for chained roles

BrickheadJohnny:
Roles can be requirement for other roles, and in this case those will be chained during access checks. The result of this chained access check is now returned in the `requirementAccesses` array, so from now we'll use that for the final result of an access check (or join), but we should still use the `children:access-check:jobs` array for the live data which we show inside the join modal

**TODO before merging the PR:**
- [x] revert `.env` changes


src/components/[guild]/JoinModal/hooks/useActiveMembershipUpdate.tsx
```diff
@@ -69,7 +80,8 @@ const useActiveMembershipUpdate = ({
                 prev?.joinedAt || res?.done ? new Date().toISOString() : null,
               roles: [
                 ...(prev?.roles?.filter(
-                  (role) => !Object.keys(byRoleId).includes(role.roleId.toString())
+                  (role) =>
+                    !Object.keys(reqJobsByRoleId).includes(role.roleId.toString())
```
Comment (cs-balazs) [inline src/components/[guild]/JoinModal/hooks/useActiveMembershipUpdate.tsx:84]: Not related to this PR, I was just wondering if doing `!(role.roleId.toString() in reqJobsByRoleId)` would be the same without having to construct the keys array, and having to iterate over it

cs-balazs (approved):
🚀

src/components/[guild]/JoinModal/utils/mapAccessJobState.ts
```diff
@@ -50,14 +50,23 @@ const mapAccessJobState = (progress: JoinJob, isLoading: boolean) => {
...
       : null
 
-  const requirements = progress["children:access-check:jobs"]
+  /**
+   * After the "access-check" step, we can use progress.requirementAccesses for live
+   * data, since that includes the final results (even if the guild has "chained"
+   * roles)
+   */
+  const requirementAccesses =
+    !!progress["completed-queue"] &&
```
Comment (dovalid) [inline src/components/[guild]/JoinModal/utils/mapAccessJobState.ts:59]: This was a little bit difficult to read, I'd suggest to move these first two lines to an isAccessCheckComplete variable so it's self explanatory what we're doing here. Other than this, everything else looks good, the comments are great, approved! (I haven't tested but saw you've done a thorough testing in DC so that's great too)

dovalid (approved):

src/components/[guild]/JoinModal/utils/mapAccessJobState.ts
Comment (BrickheadJohnny) [reply src/components/[guild]/JoinModal/utils/mapAccessJobState.ts:59]: Good point, refactored in 565612db513c894ceff0e381005d297cc9a49406!

src/components/[guild]/JoinModal/hooks/useActiveMembershipUpdate.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/JoinModal/hooks/useActiveMembershipUpdate.tsx]: According to a quick TS playground test, it would be the same, so changed it in https://github.com/guildxyz/guild.xyz/pull/1147/commits/ca21d263e135adf27a34243f8163c2a04a92a464 ;)

PR #1149, opened 2024-04-08
[GUILD-2695] - Allowlist with IPFS file

cs-balazs:
[Linear](https://linear.app/guildxyz/issue/GUILD-2695/solution-for-creatingediting-big-allowlists) - [Discord](https://discord.com/channels/697041998728659035/1221827375910621286/1225125600629489704)

- Description on BE PR: https://github.com/guildxyz/guild-backend/pull/1389
- The `Textarea` diff in `AllowlistFormInputs` is just indentation because of the `Collapse`


src/components/common/Button.tsx
```diff
@@ -45,7 +45,12 @@ const Button = forwardRef(
       )
 
     return (
-      <ChakraButton ref={ref} isLoading={isLoading} {...rest}>
+      <ChakraButton
+        ref={ref}
+        isLoading={isLoading}
+        loadingText={loadingText}
```
BrickheadJohnny, line 51: Is this actually needed? We should display `loadingTest` in line 63 already 👀

src/requirements/Allowlist/AllowlistRequirement.tsx
```diff
@@ -48,7 +48,18 @@ const AllowlistRequirement = ({ ...rest }: RequirementProps): JSX.Element => {
       {hideAllowlist ? (
         `${isEmail ? "email " : ""}allowlist`
       ) : (
-        <Button variant="link" rightIcon={<ArrowSquareIn />} onClick={onOpen}>
+        <Button
+          variant="link"
+          rightIcon={<ArrowSquareIn />}
+          {...("ipfsHash" in requirement.data
```
BrickheadJohnny, line 54: I wonder if we should dynamically load the `SearchableVirtualListModal` too only in case when we don't use this new type of allowlist? Just to optimize this requirement even more 🔥

src/requirements/Allowlist/components/AllowlistFormInputs.tsx
```diff
@@ -49,102 +57,152 @@ export default function AllowlistFormInputs({
...
+            />
+          }
+        >
           <input {...getInputProps()} hidden />
           <Text as="span" display="block" maxW={44} noOfLines={1}>
             {isDragActive ? "Drop the file here" : "Choose .txt/.csv"}
           </Text>
         </Button>
 
+        {typeof uploadedLength === "number" && (
```
BrickheadJohnny, line 138: Consider using a `Collapse` here 😉

BrickheadJohnny (commented):
Don't forget to revert the `.env` change!  
Also I left a couple of minor comments 😉

src/components/common/Button.tsx
cs-balazs: Jeje, without it, it doesn't show any loading text, just the spinner. It makes sense looking at the Chakra example: https://chakra-ui.com/docs/components/button/usage#button-loading-state
Maybe this behavior changed with a Chakra version? 🤔

src/requirements/Allowlist/AllowlistRequirement.tsx
cs-balazs: Great idea! Changed in https://github.com/guildxyz/guild.xyz/pull/1149/commits/2a2c753cdecda617b6211b0d3ad0f44487a42ad0, seems to work well 👍

src/requirements/Allowlist/components/AllowlistFormInputs.tsx
cs-balazs: Nice catch! Added in https://github.com/guildxyz/guild.xyz/pull/1149/commits/945a52fa9ab6b60b4de70e6711c3b37490ba8ed0, looks a lot smoother this way :monkayay:

src/components/common/Button.tsx
BrickheadJohnny: hmm, yeah, interesting, but it's fine then, thanks! 💯

BrickheadJohnny (approved):
🚀

PR #1150, opened 2024-04-08
fix(useAccount): use `status` instead of `isConnected`

BrickheadJohnny:
When using the `useAccount` hook, `isConnected` is `true` even if `status` is not `connected` & at that state we don't have access to `walletClient` either, so this caused some pretty strange errors on Guild.

This PR aims to fix this issue by using the `status` variable in the relevant places.

I also noticed that we still use the `isConnected` prop from wagmi in some places where we should actually check both EVM and Fuel connections, so I fixed those too.


cs-balazs (approved):
Beautiful 🚀 Can't wait to see those error events drop 💯

PR #1156, opened 2024-04-10
fix(chainsOfAddressWithDeployedContract): save response in localstorage

BrickheadJohnny:


src/hooks/useSubmit/useSubmit.ts
```diff
@@ -316,12 +315,24 @@ export const fuelSign = async ({
   return [payload, { params, sig }]
 }
 
-const chainsOfAddressWithDeployedContract = (address: `0x${string}`) =>
-  Promise.all(
+const chainsOfAddressWithDeployedContract = async (
+  address: `0x${string}`
+): Promise<Chain[]> => {
+  const LOCALSTORAGE_KEY = "chainsWithByteCode"
```
cs-balazs, line 321: Shouldn't this key include the address?

BrickheadJohnny: Whoops, you're right! Fixed in 38ffbdaed2d4dbcec15ef6e38149abcaeadb541c

cs-balazs (approved):
Lets goo 🚀

PR #1157, opened 2024-04-10
refactor: simplify Delegate functionality

BrickheadJohnny:
- Removed Delegate button from the wallet selector modal
- Removed Delegate popover (top right corner, around the account button)
- Checking user's delegations only on the currently active chain in the account modal


cs-balazs (approved):
Awesome PR as always 👑 
Works great 🚀
Note: Bruno had a copy change suggestion regarding "unlinked" in the delegate button 😉

BrickheadJohnny:
> Awesome PR as always 👑 Works great 🚀 Note: Bruno had a copy change suggestion regarding "unlinked" in the delegate button 😉

Good point, fixed in my latest commit, thanks! LFG 🚀

PR #1167, opened 2024-04-16
refactor: improve the link address flow

BrickheadJohnny:
Move the modal with the helper video to the `Web3ConnectionManager` (so we only render it once), and also showing it only when the user already picked a connector. We also don't trigger an EVM wallet connection when the user starts the link address flow, because it was confusing for Fuel users.


cs-balazs (approved):
Fantastic! 🚀 
Works like a charm, and the code is much cleaner, very nice job 💯

dovalid (approved):

PR #1168, opened 2024-04-16
feat(AddRoleReward): show settings when selecting existing platforms

dovalid:


src/platforms/Discord/DiscordCardSettings/DiscordCardSettings.tsx
```diff
@@ -1,67 +1,5 @@
...
-            <Button colorScheme="green" onClick={onClose}>
-              Done
-            </Button>
-          </ModalFooter>
-        </ModalContent>
-      </Modal>
-    </Flex>
-  )
-}
+const DiscordCardSettings = (): JSX.Element => <RoleToManage />
```
BrickheadJohnny, line 3: At this point maybe we could get rid of this wrapper component, and just pass `RoleToManage` as the `cardSettingsComponent`?

src/components/[guild]/RolePlatforms/components/EditRolePlatformButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+        size="xl"
+        closeOnEsc={!hasError}
+        closeOnOverlayClick={!hasError}
+      >
+        <ModalOverlay />
+        <ModalContent ref={modalContentRef}>
+          <ModalHeader>{`${rewardName} reward settings`}</ModalHeader>
+          <ModalBody>
+            <VStack spacing={8} alignItems="start">
+              <SettingsComponent index={index} />
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RolePlatforms/components/EditRolePlatformButton.tsx:64]: Seems like we don't need that `index` prop here 😉

src/components/[guild]/RolePlatforms/components/EditRolePlatformButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const {
+    index,
+    guildPlatform: { platformId },
+  } = useRolePlatform()
+
+  const { errors } = useFormState()
+  const hasError = !!errors?.rolePlatforms?.[index]
+
+  const rewardName =
+    (rolePlatform as RolePlatform).guildPlatform?.platformGuildName ??
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RolePlatforms/components/EditRolePlatformButton.tsx:37]: We could type props so we wouldn't need this typecast here

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/SelectExistingPlatform.tsx
```diff
@@ -52,6 +55,7 @@ const SelectExistingPlatform = ({ onClose, onSelect }: Props) => {
           if (!platformData) return null
 
           const useCardProps = platformData.cardPropsHook
+          const cardSettingsComponent = platformData.cardSettingsComponent
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/SelectExistingPlatform.tsx:58]: nit: we could maybe just destructure it in one line

```ts
const { cardPropsHook, cardSettingsComponent } = platformData
```

BrickheadJohnny (commented):
Left a couple of minor suggestions, but looks good to me overall! 🚀

src/platforms/Points/usePointsCardProps.tsx
```diff
@@ -9,7 +9,9 @@ const usePointsCardProps = (guildPlatform: GuildPlatformWithOptionalId) => {
   return {
     type: "POINTS" as PlatformName,
     image: imageUrl || <Star />,
-    name: `Get ${rolePlatform?.platformRoleData?.score} ${name || "points"}`,
+    name: `Get ${(rolePlatform?.platformRoleData?.score ?? "some") || 0} ${
```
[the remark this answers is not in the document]
dovalid: We actually need (when it's saved with no value, it's an empty string and it's better to display 0 then "some" in this case), I've added a comment about it

src/platforms/Discord/DiscordCardSettings/DiscordCardSettings.tsx
dovalid: Good point, I've left it in case we'd add new settings to Discord, but that probably won't happen so simplified it now!

PR #1170, opened 2024-04-17
feat: use Bugsnag for error monitoring

BrickheadJohnny:
I only added the error handler to our top level error boundary for testing. If we decide to use this tool, we should probably migrate the existing posthog error events to bugsnag too!


package.json
```diff
@@ -80,8 +81,9 @@
     "slugify": "^1.6.5",
     "swr": "^2.2.4",
     "uuidv7": "^0.6.3",
-    "viem": "^2.9.6",
-    "wagmi": "^2.5.17",
+    "viem": "2.9.6",
```
BrickheadJohnny, line 84: encountered a type error in newer versions, so I fixed the versions for these packages

package.json
```diff
@@ -80,8 +81,9 @@
...
+    "wagmi": "2.5.17",
+    "webpack-bugsnag-plugins": "^1.8.0",
```
cs-balazs, line 86: I'm wondering if this can be a `devDependency` 🤔

src/bugsnag.ts
```diff
@@ -0,0 +1,19 @@
...
+    window.location.host !== "guild.xyz"
+  )
+    return
+
+  Bugsnag.start({
+    apiKey: process.env.NEXT_PUBLIC_BUGSNAG_KEY,
+    plugins: [],
+    endpoints: {
+      notify: "/api/bugsnag/notify",
+      sessions: "/api/bugsnag/sessions",
```
cs-balazs, line 16: Don't we need rewrite rules for these?

cs-balazs (commented):
Nice, also, we shouldn't forget to add a domain whitelist on the BugSnag dashboard :ayy:

src/bugsnag.ts
cs-balazs: Nvm, found them 😄

cs-balazs (approved):
Great! I can't wait to see those error events 💯

package.json
BrickheadJohnny: Good point, seems like it'll work that way too, so reinstalled it as a `devDependency`!

PR #1174, opened 2024-04-18
fix(useOAuthResultToast): remove double toast & query param interpolation errors

cs-balazs:
- Issue 1: The toast is shown twice on OAuth errors, this is fixed by removing `toast` from the dependency array
- Issue 2: The `replace` call produces next js errors, which are coming to [BugSnag in a large quantity](https://app.bugsnag.com/guild-1/guild-dot-xyz/errors/6620cc8205e2280008de05f7?filters%5Bevent.unhandled%5D=true&filters%5Bevent.since%5D=30d&filters%5Bapp.release_stage%5D=production&filters%5Berror.status%5D=for_review)

### DIFF
--- a/src/hooks/useOAuthResultToast.ts
+++ b/src/hooks/useOAuthResultToast.ts
@@ -8,24 +8,31 @@ import useToast from "./useToast"
 
 export default function useOAuthResultToast() {
   const toast = useToast()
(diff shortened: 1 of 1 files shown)


BrickheadJohnny (approved):
👑 🚀

PR #1175, opened 2024-04-19
fix(waas): use non-dyamic import

cs-balazs:
- We have quite a few `Failed to load chunk` errors, this PR aims to fix that
- The dynamic import itself was a workaround for this build error:
  - `TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module ".../node_modules/@coinbase/waas-sdk-web/dist/versions.json" needs an import attribute of type "json"`
- This PR introduces a different workaround, so we can import non-dynamically
  - Adds a `fixWaas.js` script, which replaces the `varsion.json` imports of the relevant files with the correct type assertion

~~Draft until tested on preview~~ Tested it, the build is successful, and functionality is working

### DIFF
--- a/fixWaas.js
+++ b/fixWaas.js
@@ -1,0 +1,28 @@
+const { readFileSync, writeFileSync } = require("fs")
+
+const basePath = "./node_modules/@coinbase/waas-sdk-web/dist"
+
+const paths = [
+  `${basePath}/mpc.js`,
+  `${basePath}/waas.js`,
+  `${basePath}/perf.js`,
+  `${basePath}/bugsnag.js`,
+  `${basePath}/core/http.js`,
+]
+console.log(`[fixWaas] - ${paths.length} files will be processed`)
+
+for (const path of paths) {
+  console.log("\n[fixWaas] - Processing:", path)
+
+  try {
+    const content = readFileSync(path).toString()
+    const newContent = content
+      .replace(`versions.json";`, `versions.json" assert { type: "json" };`)
+      .replace(`versions.json';`, `versions.json' assert { type: "json" };`)
+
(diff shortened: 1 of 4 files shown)


fixWaas.js
```diff
@@ -0,0 +1,28 @@
...
+console.log(`[fixWaas] - ${paths.length} files will be processed`)
+
+for (const path of paths) {
+  console.log("\n[fixWaas] - Processing:", path)
+
+  try {
+    const content = readFileSync(path).toString()
+    const newContent = content
+      .replace(`versions.json";`, `versions.json" assert { type: "json" };`)
+      .replace(`versions.json';`, `versions.json' assert { type: "json" };`)
```
BrickheadJohnny, line 21: Is this duplication intentional? 👀

fixWaas.js
```diff
@@ -0,0 +1,28 @@
+const { readFileSync, writeFileSync } = require("fs")
```
BrickheadJohnny, line 1: Consider adding a short comment to this file which clarifies why do we need this 😉

BrickheadJohnny (approved):
Left one question & one suggestion, but the PR looks good already, so approving it! 👑 🚀

fixWaas.js
cs-balazs: There is a slight difference, one of the source files uses ', the rest use " :ayy:

fixWaas.js
cs-balazs: Good point! Added in https://github.com/guildxyz/guild.xyz/pull/1175/commits/2f1ec19b5e406a39c759186f27c9365a507bdbe5 :ayy:

PR #1180, opened 2024-04-23
[GUILD-2867] - Discord setup PostHog events

cs-balazs:
- Added events in the general DC add flow, plus some extra events, which are specific to guild creation, and reward addition to an existing guild
- Session recordings are started when the user initially selects discord in either flow. Not sure if these recordings are sampled by default
- One of the events is fired in a `useEffect` in `NotAdminError`. It seemed like a simple way to do it, but we might want to consider a refactor, let me know what you think
- Some events are duplicated with different names. This is because I'm not sure, which events are used, so didn't want to change the names, but wanted to stick to the same prefix for all the relevant events. But now I'm having second thoughts if it is worth it, so I think we might want to just stick to the existing events

### DIFF
--- a/src/components/[guild]/AddRewardButton/AddRewardButton.tsx
+++ b/src/components/[guild]/AddRewardButton/AddRewardButton.tsx
@@ -13,6 +13,7 @@ import {
   useColorModeValue,
   useDisclosure,
 } from "@chakra-ui/react"
+import { usePostHogContext } from "components/_app/PostHogProvider"
 import Button from "components/common/Button"
 import DiscardAlert from "components/common/DiscardAlert"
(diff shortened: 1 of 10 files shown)


BrickheadJohnny (approved):
Great PR! 🎉 

> Not sure if these recordings are sampled by default

They aren't, we should set that inside PostHog -> Settings -> Session replay 😉  And actually session replay is disabled right now, so it won't start even if we start it programatically - we should turn them on in our dashboard!

Regarding `NotAdminError`: we could actually fire those events in SWR's `onSuccess`, but it's more logical the way you did it, so I don't really have a strong opinion on this one, hehe 😄

PR #1181, opened 2024-04-24
[GUILD-2900] - Fix add bot button

cs-balazs:
- With some experimentation, revalidating the server data with `refreshInterval: 2000` seemed inconsistent, we can't guarantee that revalidation happens, when we want it to
  - Instead, I added an `onClose` callback to `usePopupWindow` which fires when the popup is closed
  - Note that this only works if we have a valid `windowInstance` object. I don't think this is a huge issue, since we used this same object for telling if the popup is open anyway
- Added `shouldRetryOnError: false` to `useServerData`'s SWR call, it seemed unnecessary to revalidate if this request fails
- Also, I've seen that gateables is also revalidated every 10 seconds, not sure why that is needed, the only thing I can think of is to handle if a server is created during the guild creation flow, which I think is an unlikely case 🤔

### DIFF
--- a/src/components/common/DiscordGuildSetup/components/DCServerCard.tsx
+++ b/src/components/common/DiscordGuildSetup/components/DCServerCard.tsx
@@ -5,6 +5,7 @@ import CardMotionWrapper from "components/common/CardMotionWrapper"
 import OptionCard from "components/common/OptionCard"
 import usePopupWindow from "hooks/usePopupWindow"
 import useServerData from "hooks/useServerData"
+import useSubmit from "hooks/useSubmit"
 import Link from "next/link"
 import { ArrowSquareIn } from "phosphor-react"
 import usePlatformUsageInfo from "platforms/hooks/usePlatformUsageInfo"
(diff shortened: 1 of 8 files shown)


src/components/common/DiscordGuildSetup/components/DCServerCard.tsx
```diff
@@ -25,17 +26,20 @@ const DCServerCard = ({ serverData, onSelect, onCancel }: Props): JSX.Element =>
...
-  } = useServerData(serverData.id, {
-    swrOptions: {
-      refreshInterval: !!activeAddBotPopup ? 2000 : 0,
-      refreshWhenHidden: true,
-    },
+    mutate,
+  } = useServerData(serverData.id)
+
+  // useSubmit is used here for the loading state
+  const { onSubmit: onCheckBot, isLoading: isCheckingBot } = useSubmit(async () => {
```
BrickheadJohnny, line 41: Can't we use `isValidating` from `useServerData` for this? 👀

BrickheadJohnny (commented):
Great changes, left 1 minor suggestion/question! 😉 

>  I've seen that gateables is also revalidated every 10 seconds, not sure why that is needed, the only thing I can think of is to handle if a server is created during the guild creation flow, which I think is an unlikely case

Yup, exactly, I think we can remove that!

PR #1182, opened 2024-04-24
feat: rework the Collect NFT page

BrickheadJohnny:
This PR aims to fix several problems on the Collect NFT page:

- the biggest change is that we use our own API key during ISR on this page, so we can properly determine if the reward actually doesn't exist (in this case, we can return a 404 response), or it's a hidden/private reward
- added a new serverless endpoint (`/api/nft/[chain]/[address]`): we fetch the immutable onchain data for the NFT on this endpoint, and cache it for a very long time, so we won't need to re-fetch it on the client & we also pre-populate the SWR cache with it during ISR, so we'll be able to load render the page data on client side much faster
- fixed `RoleRequirements`: we only show a skeleton loader when the requirements are actually loading
- fixed the SWR mutation in `useCreateNft`, it probably didn't work at all before (it mutated a non-existent cache key)

**TODO**

- [x] add the `GUILD_API_KEY` env var on Vercel

### DIFF
--- a/src/pages/[guild]/collect/[chain]/[address].tsx
+++ b/src/pages/[guild]/collect/[chain]/[address].tsx
@@ -47,6 +47,7 @@ type Props = {
   guildPlatformId: number
   roleId: number
   rolePlatformId: number
+  fallbackImage?: string
   fallback: { [x: string]: Guild }
 }
 
@@ -61,6 +62,7 @@ const CollectNftPageContent = ({
   address,
   guildPlatformId,
   roleId,
+  fallbackImage,
 }: Omit<Props, "urlName" | "fallback">) => {
   const { theme, guildPlatforms, roles } = useGuild()
   const { isAdmin } = useGuildPermission()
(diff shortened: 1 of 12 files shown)


src/pages/[guild]/collect/[chain]/[address].tsx
```diff
@@ -23,210 +24,184 @@ import useNftDetails from "components/[guild]/collect/hooks/useNftDetails"
...
-        captureEvent("ErrorBoundary catched error", {
-          page: "[guild]/collect/[chain]/[address]",
-          guild: urlName,
-          nftAddress: address,
-          error,
-          info,
-        })
-      }}
-      fallback={<ErrorPage />}
+    <Layout
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/collect/[chain]/[address].tsx:78]: Removed the ErrorBoundary, BugSnag will catch issues from now 😉

src/pages/[guild]/collect/[chain]/[address].tsx
```diff
@@ -236,22 +218,75 @@ const getStaticProps: GetStaticProps<Props> = async ({ params }) => {
...
-  const endpoint = `/v2/guilds/guild-page/${urlName}`
-  const guild: Guild = await fetcher(endpoint).catch((_) => ({}))
+  const guildPageEndpoint = `/v2/guilds/guild-page/${urlName}`
 
-  if (!guild?.id)
+  let publicGuild: Guild, guild: Guild
+  try {
+    ;[publicGuild, guild] = await Promise.all([
+      fetcher(guildPageEndpoint),
+      fetcher(
```
Comment (cs-balazs) [inline src/pages/[guild]/collect/[chain]/[address].tsx:227]: I'm wondering if later on we should add an api-key protected endpoint specifically for this usage, so we don't request the whole guild-page 🤔

cs-balazs (approved):
Amazing ✨ 
Left just one thought, we might have brainstormed about that IRL, if I remember correctly, just a possible future improvement to avoid fetching the guild-page twice :ayy:

PR #1184, opened 2024-04-25
logging: posthog events added

FBalint:


src/components/[guild]/AddRewardButton/useCreateTokenReward.ts
```diff
@@ -63,10 +70,18 @@ const useCreateReqBasedTokenReward = ({
...
+      captureEvent(
+        "Failed to add token reward (multiple roles selected)",
+        postHogOptions
+      )
       showErrorToast("Dynamic token rewards can be added to only one role at most.")
       return
     }
     if (data.requirements.length > 1) {
+      captureEvent(
+        "Failed to add token reward (multiple requirements set when adding to existing role)",
```
Comment (BrickheadJohnny) [inline src/components/[guild]/AddRewardButton/useCreateTokenReward.ts:82]: In order to keep these event names short, we could just call them "Failed to add token reward" & pass the reason (e.g. "multiple requirements set when adding to existing role) in an event property ;)

src/components/[guild]/AddRewardButton/useCreateTokenReward.ts
```diff
@@ -82,6 +97,11 @@ const useCreateReqBasedTokenReward = ({
          * needed in the reward's rolePlatform's dynamicData field.
          */
 
+        captureEvent("createTokenReward(AddToExistingRole) Requirement created", {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/AddRewardButton/useCreateTokenReward.ts:100]: Not sure if we need this one

src/components/[guild]/AddRewardButton/useCreateTokenReward.ts
```diff
@@ -117,9 +137,23 @@ const useCreateReqBasedTokenReward = ({
           { revalidate: false }
         )
 
-        onAddRewardSubmit(rewardSubmitData).then(() => triggerMembershipUpdate())
+        onAddRewardSubmit(rewardSubmitData).then(() => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/AddRewardButton/useCreateTokenReward.ts:140]: Let's put it in the hook's `onSuccess` param instead

src/components/[guild]/AddRewardButton/useCreateTokenReward.ts
```diff
@@ -138,6 +172,20 @@ const useCreateReqBasedTokenReward = ({
       rolePlatforms: [],
       logic: "AND",
       guildId,
+    }).catch((error) => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/AddRewardButton/useCreateTokenReward.ts:175]: Consider using `useCreateRole`'s `onError` instead

src/components/[guild]/AddRewardButton/useCreateTokenReward.ts
```diff
@@ -165,7 +213,21 @@ const useCreateReqBasedTokenReward = ({
       saveAs,
       createdRole.id
     )
-    await onAddRewardSubmit(rewardSubmitData)
+    const createdReward = await onAddRewardSubmit(rewardSubmitData).catch(
+      (error) => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/AddRewardButton/useCreateTokenReward.ts:217]: I think we could also use the hook's `onError` prop here

src/platforms/Token/EditTokenModal.tsx
```diff
@@ -106,6 +107,12 @@ const EditTokenModal = ({
   const { data: requirements } = useRequirements(role.id)
   const snapshotRequirement = requirements?.find((req) => !!req?.data?.snapshot)
 
+  const { captureEvent } = usePostHogContext()
```
BrickheadJohnny, line 110: If we won't allow users to add this reward in the edit role drawer, we don't need events here ;)

src/platforms/Token/hooks/useCollectToken.tsx
```diff
@@ -52,7 +59,17 @@ const useCollectToken = (
         method: "POST",
         body: {},
       },
-    ])
+    ]).catch((error) => {
```
BrickheadJohnny, line 62: We could use the hook's `onError` here too

src/platforms/Token/EditTokenModal.tsx
FBalint: But they can be edited through the access hub card, are you sure?

src/platforms/Token/hooks/useCollectToken.tsx
FBalint: This one is specifically for the claim endpoints errors. Does `useFetcherWithSign` have an `onError` function as well?

BrickheadJohnny: Oh, my bad, then it's fine this way!

src/platforms/Token/EditTokenModal.tsx
BrickheadJohnny: Oh, I forgot about that. You're right, then let's leave these events here! 👌

PR #1191, opened 2024-05-02
fix(MultiPlatformSelectButton): socialLinks field reset

cs-balazs:
- Adding a Twitter link, then removing it results in a validation error in the guild creation request


BrickheadJohnny (approved):
lfg 🚀

PR #1194, opened 2024-05-03
refactor(useCreateRequirement): submit function receives roleId

FBalint:
Removed the `useCreateRequirementForRole` hook as it's no longer necessary. Originally, it was used when adding token rewards to existing roles, but since this feature has been deferred, we can safely eliminate this portion of the code.


BrickheadJohnny (approved):
Nice :feelsgoodman: 🚀

PR #1198, opened 2024-05-06
refactor: extend ControlledNumberInput

FBalint:
…NumberInput


src/requirements/WalletActivity/components/ControlledNumberInput.tsx
```diff
@@ -34,11 +50,96 @@ const ControlledNumberInput = ({
     },
   })
 
+  const decimalsLimit = decimalsMax < MAX_DECIMALS ? decimalsMax : MAX_DECIMALS
```
BrickheadJohnny, line 53: nit: `Math.min(decimalsMax, MAX_DECIMALS)` might be a cleaner solution, wdyt?

src/requirements/WalletActivity/components/ControlledNumberInput.tsx
```diff
@@ -34,11 +50,96 @@ const ControlledNumberInput = ({
...
+
+  /**
+   * Makes the step size adapt to the precision. E.g., if the entered number is
+   * 1.001, the step size will be set to 0.001, therefore the next step up will be
+   * 1.002.
+   */
+  const updateStepSize = useCallback(
```
BrickheadJohnny, line 60: I think we could just use a `useMemo` here, so we wouldn't need a `useState`, a function wrapped in `useCallback` & a `useEffect` just to handle the `stepSize`

BrickheadJohnny (changes requested):
Great work, but I left a couple of suggestions. Let me know if you have any questions! 👍

FBalint:
> Great work, but I left a couple of suggestions. Let me know if you have any questions! 👍

Thanks, the subform was a great idea! It also helped me realise that the whole `ConversionInput` component could be made way more generic, as you can see in newer commits.

Also, not completely related to this PR, but whilst working on this, I noticed a pattern for finding certain guildPlatforms. I extracted this logic to a simple hook called `useGuildPlatform`, and used it where I found appropriate.

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddTokenPanel/components/TokenAmountStep.tsx
```diff
@@ -134,7 +137,7 @@ const TokenAmountStep = ({ onContinue }: { onContinue: () => void }) => {
           ) : (
             <StaticAmount />
           )} */}
-          <DynamicAmount />
+          <DynamicAmount tokenData={tokenData} />
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddTokenPanel/components/TokenAmountStep.tsx:140]: I'd just call `useTokenData` inside this component ;)

BrickheadJohnny (approved):
Left 1 more short comment, otherwise LGTM!

PR #1199, opened 2024-05-06
[GUILD-2863] - usePinata optimization

cs-balazs:
[Linear](https://linear.app/guildxyz/issue/GUILD-2863/fix-usepinata-hook)

- Tested the changes by adding a `useEffect(() => console.log("onUpdate", onUpdate), [onUpdate])` at the end of `usePinata` before the `return`. Then visiting the `/create-guild` page, and on the `BasicInfo` section, I've made some blur-s on the inputs. The blurs caused both of the `usePinata` hooks to log the message again (plus there were an initial ~8 logs)
- I think the reason is that the form's context changed and that we used anonymous functions without `useCallback`
- Changes in this PR:
  - `useSubmit`: Wrapped the returned `onSubmit` function in a useCallback - [a8dc2a0](https://github.com/guildxyz/guild.xyz/commit/a8dc2a0c4e0e920c6b210b56e40782893a557d34)
  - `usePinata`
    - Moved the submit function out of the hook - [c3a84d2](https://github.com/guildxyz/guild.xyz/commit/c3a84d274cc83f28cddcae56b28faa5fc006fe83)
    - Wrapped the `onSuccess` and `onError` callbacks in `useCallback` calls - [c3a84d2](https://github.com/guildxyz/guild.xyz/commit/c3a84d274cc83f28cddcae56b28faa5fc006fe83)
    - **Refactor**: Added `fieldToSetOnSuccess` and `fieldToSetOnError` props. If set, the hook will perform the `setValue` calls on success/error. This seemed like a good idea, because in most cases, we would just pass callbacks to set a form field, and this way we can avoid having to wrap all those callbacks in `useCallback`s - [778e9bc](https://github.com/guildxyz/guild.xyz/commit/778e9bc8bfc6cd405bf08a93e92c32c73a1602b2)
  - Made the `usePinata` call changes in the places, where the hook is used
    - `BasicInfo`: [778e9bc](https://github.com/guildxyz/guild.xyz/commit/778e9bc8bfc6cd405bf08a93e92c32c73a1602b2)
    - Rest: [f399e09](https://github.com/guildxyz/guild.xyz/commit/f399e0961885b0b17ba5de085c2d6d3563a0715d)

### DIFF
--- a/src/hooks/usePinata/usePinata.ts
+++ b/src/hooks/usePinata/usePinata.ts
@@ -1,6 +1,8 @@
 import useSubmit from "hooks/useSubmit"
 import useToast from "hooks/useToast"
 import { useCallback } from "react"
+import { useFormContext } from "react-hook-form"
+import getRandomInt from "utils/getRandomInt"
 import pinFileToIPFS, {
   PinToIPFSProps,
   PinataPinFileResponse,
@@ -9,6 +11,8 @@ import pinFileToIPFS, {
 type Props = Partial<{
   onSuccess: (data: PinataPinFileResponse) => void
   onError: (error: any) => void
+  fieldToSetOnSuccess: string
+  fieldToSetOnError: string
 }>
 
 export type Uploader = {
@@ -18,8 +22,14 @@ export type Uploader = {
 
 const submit = (ipfsProps: PinToIPFSProps) => pinFileToIPFS(ipfsProps)
 
-const usePinata = ({ onError, onSuccess }: Props = {}): Uploader => {
+const usePinata = ({
+  onError,
+  onSuccess,
+  fieldToSetOnSuccess,
+  fieldToSetOnError,
+}: Props = {}): Uploader => {
   const toast = useToast()
+  const { setValue } = useFormContext() ?? {}
 
   const wrappedOnError = useCallback(
     (error) => {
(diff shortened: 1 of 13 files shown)


src/hooks/usePinata/usePinata.ts
```diff
@@ -1,38 +1,79 @@
...
+  fieldToSetOnError: string
 }>
 
 export type Uploader = {
   onUpload: (data?: PinToIPFSProps) => void
   isUploading: boolean
 }
 
-const usePinata = ({ onError, onSuccess }: Props = {}): Uploader => {
+const submit = (ipfsProps: PinToIPFSProps) => pinFileToIPFS(ipfsProps)
```
BrickheadJohnny, line 23: Couldn't we just pass `pinFileToIPFS` to `useSubmit` below? 👀

BrickheadJohnny (approved):
Awesome refactor, left one tiny suggestion, but approving it! 🚀

src/hooks/usePinata/usePinata.ts
cs-balazs: Oh, that is a nice catch, we should be able to pass that, let me try that out!

cs-balazs: It seems to work as expected! Changed in https://github.com/guildxyz/guild.xyz/pull/1199/commits/3d35dbb3895eb39fb6094c7da43008b896eee0cf

PR #1201, opened 2024-05-07
feat(NFT reward): new availability related features

BrickheadJohnny:
- Added new `NftAvailabilityTags` for NFT rewards - but we still render conditionally the "legacy" availability tags (from now we'll store capacity inside the contract instead of our backend)
- Added support for the new claim function (but kept the old one too for backward compatibility)
- Added a new amount picker component to the collect page, with ranges and fun icons
- Updated the collectors section (added number of mints & range icons) & serverless endpoint (returning balances now)
- Created a `useGuildRewardNftBalanceByUserId` hook, which checks the user's NFT balance by user ID instead of address
- Updated `TransactionStatusContext`: added `assetAmount`, so we can show the number of minted NFTs in the success modal / on the success toast


src/platforms/ContractCall/ContractCallReward.tsx
```diff
@@ -61,7 +67,14 @@ const ContractCallReward = ({
         </>
       }
     >
-      <AvailabilityTags rolePlatform={platform} />
+      {contractCallFunction === ContractCallFunction.DEPRECATED_SIMPLE_CLAIM ? (
```
dovalid, line 70: Maybe we could abstract this condition into NftAvailabilityTags, so we wouldn't need to repeat it here and useContractCallCardProps? Just use NftAvailabilityTags as it'll remain, so the legacy support in it is just an implementation detail that we'll be able to remove later

BrickheadJohnny: Good point, refactored this in 2a773621dd6dfbb5b1cfff124104b60404b35d26

dovalid (approved):

PR #1204, opened 2024-05-08
[GUILD-2917] - Allowlist request optimization with searching

cs-balazs:
- We only get at maximum 10 addresses in a request
- We can pass a `search` query param to search in the allowlist
- Added a `useRequirement` hook to manage this logic


BrickheadJohnny (approved):
Great changes! 🚀

Issue #1202, opened 2024-05-07
Excessive full page rerendering caused by `Layout` component


dominik-stumpf:
### Describe the bug

> [!NOTE] 
> `Layout` refers to component at [`/src/components/common/Layout/Layout.tsx`](https://github.com/guildxyz/guild.xyz/blob/da0cf25bdab18cf70c34bba0349420c60eb51fb3/src/components/common/Layout/Layout.tsx)

There is an ongoing issue with page route `/create-guild` at the moment where full sized page rerenders are occurring when updating form fields or progressing through the guild creation process as shown in the reproduction video.

As a result this introduces performance issues where form validations are happening, severely damaging the end user experience of newcomers and regular users, especially the ones with low-end devices.

After investigating, I found that problem is located in the widely used `Layout` component which could make this issue touch other areas of the website and overall this is not a good practice in react. The performance problem will grow as this component does making the site less and less usable.

### To Reproduce

https://github.com/guildxyz/guild.xyz/assets/122315398/25c65e8b-da0a-44a4-a19b-138931de3385

React dev tools highlights with the rectangles how areas like the `Header` is being rerendered on every keystroke, interaction abundantly.

### Expected Behavior

Should strictly rerender only components that use the relevant input value.

### Which browser are you using? (if relevant)

Browser irrelevant.

dominik-stumpf:
If you don't mind I would like to open a discussion or PR about this issue soon.

BrickheadJohnny:
Hey, thanks for the detailed report. We should definitely fix it, feel free to open a PR if you wish @dominik-stumpf!

BrickheadJohnny:
Issue fixed in #1205

PR #1205, opened 2024-05-08
perf: throttle text input fields

dominik-stumpf:
Performance improvement for the second stage of guild creation process, motivated by #1202.

PR Result:

https://github.com/guildxyz/guild.xyz/assets/122315398/ceb490b0-1ab4-479a-b98a-74aedf067981

### DIFF
--- a/src/components/create-guild/BasicInfo/components/ContactInfo.tsx
+++ b/src/components/create-guild/BasicInfo/components/ContactInfo.tsx
@@ -14,6 +14,7 @@ import {
 import Button from "components/common/Button"
 import FormErrorMessage from "components/common/FormErrorMessage"
 import StyledSelect from "components/common/StyledSelect"
+import { useThrottledRegister } from "components/create-guild/hooks/useThrottledRegister"
 import { ArrowSquareOut, Plus, TrashSimple } from "phosphor-react"
 import { Controller, useFieldArray, useFormContext } from "react-hook-form"
 import { GuildFormType, SelectOption } from "types"
@@ -36,6 +37,7 @@ const ContactInfo = ({ showAddButton = true }: Props): JSX.Element => {
     resetField,
     formState: { errors },
   } = useFormContext<GuildFormType>()
+  const { throttledRegister } = useThrottledRegister(register, 300)
 
   const { fields, append, remove } = useFieldArray({
     control,
@@ -92,7 +94,7 @@ const ContactInfo = ({ showAddButton = true }: Props): JSX.Element => {
                         ? `E-mail address`
                         : "Phone / Telegram username"
                     }
-                    {...register(`contacts.${index}.contact`, {
+                    {...throttledRegister(`contacts.${index}.contact`, {
                       required: "This field is required",
                       pattern:
                         getValues(`contacts.${index}.type`) === "EMAIL"
--- a/src/components/create-guild/Name.tsx
+++ b/src/components/create-guild/Name.tsx
@@ -1,6 +1,7 @@
 import { FormControl, Input, ResponsiveValue } from "@chakra-ui/react"
 import FormErrorMessage from "components/common/FormErrorMessage"
 import { useFormContext } from "react-hook-form"
+import { useThrottledRegister } from "./hooks/useThrottledRegister"
 
 type Props = {
   isDisabled?: boolean
@@ -17,6 +18,7 @@ const Name = ({
     register,
     formState: { errors },
   } = useFormContext()
+  const { throttledRegister } = useThrottledRegister(register, 300)
 
   return (
     <FormControl
@@ -27,7 +29,7 @@ const Name = ({
     >
       <Input
         size="lg"
-        {...register("name", {
+        {...throttledRegister("name", {
           required: "This field is required.",
           maxLength: {
             value: 50,
--- a/src/components/create-guild/hooks/useThrottledRegister.ts
+++ b/src/components/create-guild/hooks/useThrottledRegister.ts
@@ -1,0 +1,33 @@
+import { useRef } from "react"
+import type { ChangeHandler, FieldValues, UseFormRegister } from "react-hook-form"
+
+export function useThrottledRegister(
+  register: UseFormRegister<FieldValues>,
+  delayMs: number
+) {
+  const timeoutId = useRef<number | undefined>()
+
+  function throttledRegister(
+    ...registerArgs: Parameters<UseFormRegister<FieldValues>>
+  ): ReturnType<UseFormRegister<FieldValues>> {
+    const registerHandlers = register(...registerArgs)
+    const defaultOnChange = registerHandlers.onChange
+
+    registerHandlers.onChange = async (
+      ...handlerArgs: Parameters<ChangeHandler>
+    ) => {
+      const startTimeout = () => {
+        timeoutId.current = window.setTimeout(() => {
+          defaultOnChange(...handlerArgs)
+        }, delayMs)
+      }
+
+      clearTimeout(timeoutId.current)
+      startTimeout()
+    }
+
+    return registerHandlers
+  }
+
+  return { throttledRegister }
+}


dominik-stumpf:
Yes, I agree on that. Throttling is meant to be a temporary solution after all. I was thinking about what could be done with the `Layout` and I believe the problem is not with the component itself, but how it is being used. Maybe I can rewrite it to avoid further misuse of it like here.

The thing is that last time I counted 61 references on `Layout` and honestly I am not too keen on changing a component that is used so much without more insight.

I could however make instead a dedicated `Layout` for the `create-guild` section that would push down state deeper as you mentioned and experiment with it without interfering with the original `Layout`. What are your thoughts on that @BrickheadJohnny?

BrickheadJohnny:
I think the component is actually used "only" 21 times in our codebase, but yeah, that's still a lot! I wonder if we could "slice up" and refactor it (~ compound component?), so we could have a `Layout`, `Layout.Header`, `Layout.Description`, and a `Layout.Footer`, and we'd just build our layouts using these? I'm pretty sure that way we'd solve these re-render issues, e.g. if the title changes, only `Layout.Header` would re-render, not the whole page.

Even if we do this refactor, we might need to wrap the `Layout.Header` in a new component, which uses `useWatch({ name: "name" })` to update its title to the guild's name on the Create guild page (instead of watching this value in the root `create-guild.tsx` component), but we'll see that later. 

As a side quest, we could also get rid of that `useIsomorphicLayoutEffect` and structure this component in a way which wouldn't require JS for setting the cover's height.

What do you think about this approach @dominik-stumpf?

dovalid:
Hey @dominik-stumpf, thanks so much for the issue and PR! I agree with Johnny that we should actually use composition for the Layout component, but also think that just debouncing the state like you did is a great quick fix! As Johnny wrote, I wouldn't alter the register function neither tho - we actually have a useDebouncedState hook, that's perfect for this usecase. If you could implement it with that, I'd be happy to merge this PR, and refactor the Layout component to use composition later 🔥

dominik-stumpf:
> Hey @dominik-stumpf, thanks so much for the issue and PR! I agree with Johnny that we should actually use composition for the Layout component, but also think that just debouncing the state like you did is a great quick fix! As Johnny wrote, I wouldn't alter the register function neither tho - we actually have a useDebouncedState hook, that's perfect for this usecase. If you could implement it with that, I'd be happy to merge this PR, and refactor the Layout component to use composition later 🔥

There is no need. I have a solution in mind for the re-render issue so I'm taking back the throttle. I've made an experiment where I pulled apart the `Layout`, and the form on `/create-guild` is way more pleasant already :)

For this PR I will keep focus on `/create-guild` because I feel like it's a crucial part of the site and I want it to be smooth and snappy. As for the solution I would like to create a component called `CompoundLayout` *as suggested* and refactor there the `Layout` properly and keep it fully compatible without altering the flow of code too much. So if all goes well I could open another PR, where the `Layout` gets replaced with `CompoundLayout` in the rest of the 20 files (`create-guild` would be the example on how use the new `Layout`).

> As a side quest, we could also get rid of that `useIsomorphicLayoutEffect` and structure this component in a way which wouldn't require JS for setting the cover's height.

It should be possible to replace it. I don't know if it will be changed here, that depends on how it affects my refactoring. It looks out of place for me as well.

dominik-stumpf:
The new `Layout` is working with the create guild, however it's not tested elsewhere. I would like to cover that in another PR. The background is now covered by CSS, but I didn't quite get why the `window.scrollY` was appended to the offset shown here:

```tsx
setBgHeight(`${rect.top + (window?.scrollY ?? 0) + backgroundOffset}px`)
```

I left out that part from the new version therefore the background height is not exactly the same. Also I couldn't decide whether we should have a failsafe feature that checks in dev mode if the compound component is constructed correctly (`doesComponentExistIn.ts` utility's purpose).

As for the create guild optimization, I looked through the guild stepper and found that there is still room for improvement beyond the `Layout`.

src/components/common/CompoundLayout/Layout.tsx
```diff
@@ -12,8 +12,8 @@ import useIsomorphicLayoutEffect from "hooks/useIsomorphicLayoutEffect"
 import Head from "next/head"
 import Image from "next/image"
 import { PropsWithChildren, ReactNode, useRef, useState } from "react"
-import Footer from "./components/Footer"
-import Header from "./components/Header"
+import { Footer } from "./components/Footer"
```
BrickheadJohnny, line 15: We usually use default exports when we export only 1 component from a file ;)

src/components/common/CompoundLayout/components/Head.tsx
```diff
@@ -0,0 +1,22 @@
...
+interface Props {
+  ogTitle?: string
+  ogDescription?: string
+  imageUrl?: string
+  title?: JSX.Element | string
+}
+
+export const Head = ({ ogTitle, ogDescription, title, imageUrl }: Props) => (
+  <NextHead>
+    <title>{`${ogTitle ?? title}`}</title>
```
BrickheadJohnny, line 12: This can be problematic when `title` is a `JSX.Element`

src/components/common/CompoundLayout/components/Root.tsx
```diff
@@ -0,0 +1,38 @@
...
+  background?: string
+  maxWidth?: string
+}>
+
+export function Root({
+  children,
+  background,
+  maxWidth = "sizes.container.lg",
+}: Props) {
+  const { colorMode } = useColorMode()
```
BrickheadJohnny, line 15: We could clean up the code a bit by using the `useColorModeValue` hook here. I know we used `useColorMode` before this refactor too, but I think we could change it in this PR.

src/components/common/CompoundLayout/components/Section.tsx
```diff
@@ -0,0 +1,28 @@
+import { Container, VStack } from "@chakra-ui/react"
+import { ReactNode } from "react"
+import { callCssVariable } from "../callCssVariable"
+import { LAYOUT_MAX_WIDTH_CSS_VAR } from "../constants"
+
+interface SectionProps {
+  children: ReactNode
+}
+
+export function MainSection({ children }: SectionProps) {
```
BrickheadJohnny, line 10: Let's make 2 different files: `MainSection` & `HeaderSection`. We already have a `Section` component, so it'd be a bit confusing to have 2 components with the same name

src/components/common/CompoundLayout/components/Root.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  return (
+    <Box
+      sx={{ [LAYOUT_MAX_WIDTH_CSS_VAR]: maxWidth }}
```
BrickheadJohnny, line 19: I have a suggestion for an alternative solution here. What if we'd remove this constant & type the `maxWidth` prop as `ChakraProps["width"]`? That way we could just use `maxWidth="container.lg"` here. We could also store the default value in a constant inside this file instead of a new one & export it, that way we wouldn't really need the `callCssVariable` util.

src/components/common/CompoundLayout/components/Content.tsx
```diff
@@ -0,0 +1,14 @@
+import { Box, BoxProps } from "@chakra-ui/react"
+import { PropsWithChildren, useRef } from "react"
+
+type Props = PropsWithChildren<Omit<BoxProps, "title">>
+
+export const Content = ({ children, ...wrapperProps }: Props) => {
+  const childrenWrapper = useRef(null)
```
BrickheadJohnny, line 7: I think this ref is now unused, so we don't really need it?

src/components/common/CompoundLayout/doesComponentExistIn.ts
```diff
@@ -0,0 +1,20 @@
+import { Children, ReactElement, ReactNode, isValidElement } from "react"
+
+export function doesComponentExistIn(
```
BrickheadJohnny, line 3: I think we don't use this function.

src/components/common/CompoundLayout/SimpleLayout.tsx
```diff
@@ -0,0 +1,50 @@
...
+import Head from "next/head"
+import { PropsWithChildren } from "react"
+import Footer from "./components/Footer"
+import Header from "./components/Header"
+
+type Props = {
+  title: string
+}
+
+const SimpleLayout = ({
```
BrickheadJohnny, line 11: I think this is an unused component, so we can remove it

BrickheadJohnny (changes requested):
Great work overall, but I left a couple of comments. Also, I'd suggest to revert the `/Layout` -> `/CompoundLayout` renames, and just rename the old layout component to `LegacyLayout` in order to keep the PR as small as possible.

Let's chat on Discord if you have any questions. 😉

src/components/common/CompoundLayout/doesComponentExistIn.ts
dominik-stumpf: I had an ongoing idea about a failsafe to check if the `Layout` is constructed properly. This function would help create this feature. Do you think we need this failsafe? @BrickheadJohnny

BrickheadJohnny: Yeah, I like the idea, but since this is one of the main components on each page, I think we'd notice any issues with it during development, so we don't need to add this extra logic to it, let's keep it simple.

dominik-stumpf:
> Great work overall, but I left a couple of comments. Also, I'd suggest to revert the `/Layout` -> `/CompoundLayout` renames, and just rename the old layout component to `LegacyLayout` in order to keep the PR as small as possible.

Well.. Renaming `Layout` to `LegacyLayout` makes new files either way, but the renaming is done and I put a deprecated marker on the old layout to discourage others from using it.

src/components/common/Layout/components/NavMenu/NavMenu.tsx
```diff
@@ -102,7 +102,7 @@ const NavMenu = (): JSX.Element => {
             <NavGroup title="Other">
               <NavButton
                 target="_blank"
-                href="https://discord.gg/KUkghUdk2G"
+                href="https://discord.gg/guildxyz"
```
BrickheadJohnny, line 105: Nope, please revert it, we don't have a pretty invite anymore & this redirects to a random server, haha

BrickheadJohnny (approved):
Seems like the `Section.tsx` file is empty, let's remove it then. Otherwise, LGTM. Let's wait for @dovalid's review too.

src/components/common/Layout/components/NavMenu/NavMenu.tsx
dominik-stumpf: Shouldn't we remove it then?

src/utils/wrapInCssVar.ts
```diff
@@ -0,0 +1,3 @@
+export function wrapInCssVar(cssVariable: `--${string}`) {
```
dovalid, line 1: I don't think this abstraction is worth it, it's more boilerplate & unnecessary complexity to use than just typing `var()` where we need it

src/components/common/Layout/constants.ts
```diff
@@ -0,0 +1 @@
+export const LAYOUT_MAX_WIDTH_CSS_VAR = '--layout-max-width'
```
dovalid, line 1: Same, we shouldn't make a JS variable if we've already made a CSS one

src/components/common/Layout/components/Account/components/AccountModal/components/SocialAccount/EmailAddress.tsx
```diff
@@ -159,10 +159,9 @@ const ConnectEmailButton = ({
         colorScheme={emails?.pending ? "orange" : rewards.EMAIL.colorScheme}
         variant={"solid"}
         size="sm"
-        isDisabled={emails?.emailAddress}
```
dovalid, line 162: Why are these changes here? If they fix something unrelated, it should be a separate PR, but also seem like they're unintentional?

dominik-stumpf: They are unintentional. Let me look into it

src/components/common/Layout/components/Head.tsx
```diff
@@ -0,0 +1,24 @@
+import NextHead from "next/head"
+
+interface Props {
+  ogTitle?: string
+  ogDescription?: string
+  imageUrl?: string
+  title?: string
```
dovalid, line 7: We don't need the title prop now, it was just used as a fallback if there's no ogTitle passed - now that it's a separate component anyway, we should make it required and remove the title prop

src/components/common/Layout/components/Content.tsx
```diff
@@ -0,0 +1,10 @@
+import { Box, BoxProps } from "@chakra-ui/react"
+import { PropsWithChildren } from "react"
+
+type Props = PropsWithChildren<Omit<BoxProps, "title">>
+
+const Content = ({ children, ...wrapperProps }: Props) => (
```
dovalid, line 6: I think we should just abstract it into MainContent, if we need it at all, but I'm not sure about that as that container is just a simple div too. I'd remove for now and add back to MainContent later when applying the new layout for other pages too, if we happen to need it

src/components/common/Layout/components/Headline.tsx
```diff
@@ -0,0 +1,54 @@
...
+  title?: JSX.Element | string
+  description?: JSX.Element
+  action?: ReactNode | undefined
+}
+
+const Headline = ({ image, title, description, action }: Props) => {
+  const { textColor } = useThemeContext()
+
+  return (
+    (image || title || description) && (
```
dovalid, line 16: We can remove this condition and make title required, it was for pages like the NFT minting page where we don't want this section at all, now we just won't include the Headline component there

src/components/common/Layout/components/Background.tsx
```diff
@@ -0,0 +1,58 @@
...
+const Background = ({
+  background,
+  backgroundImage,
+  backgroundProps,
+  backgroundOffset = 128,
+}: Props) => {
+  const colorContext = useThemeContext()
+
+  return (
+    (background || backgroundImage) && (
```
dovalid, line 23: Same here, we just won't include the component if we don't want it on a page

src/components/common/Layout/components/Background.tsx
```diff
@@ -0,0 +1,58 @@
+import { Box, BoxProps } from "@chakra-ui/react"
+import { useThemeContext } from "components/[guild]/ThemeContext"
+import Image from "next/image"
+
+type Props = {
+  image?: JSX.Element
```
dovalid, line 6: We should remove the image and imageUrl props, remove the background prefix from the other props, and handle that we either have an image or a background prop. If we have the latter, we should just add `& BoxProps` instead of the backgroundProps we have now, so they can be spread to the element

src/components/common/Layout/index.tsx
```diff
@@ -1,3 +1,77 @@
...
+   *
+   * For specifing `maxWidth` property, use `sizes.container.<size>`
+   */
+  Root,
+  Header,
+  Footer,
+  /** `NextHead` wrapper for page metadata */
+  Head,
+  /**
+   * `HeaderSection` component that displays an adjustable background
```
dovalid, line 60: Should be `Background`, right?

src/components/common/Layout/index.tsx
```diff
@@ -1,3 +1,77 @@
...
+   *
+   * Fills the area of `HeaderSection` + `backgroundOffset`
+   */
+  Background,
+  /** Main content of the page that `Layout` wraps around */
+  Content,
+  BackButton,
+  /**
+   * `HeaderSection` component that displays page title, description and other
```
dovalid, line 69: Should be `Headline`, right?

src/components/common/Layout/index.tsx
dovalid: actually, we shouldn't repeat the component names in the comments, so should just be `Displays an adjustable background, that fills the area of "HeaderSection" + "backgroundOffset"`

src/components/common/Layout/index.tsx
dovalid: same as above, should just be `Displays page title, image and description`

src/pages/create-guild.tsx
```diff
@@ -1,98 +1,28 @@
...
+import { CreateGuildContent } from "components/create-guild/CreateGuildContent"
+import { CreateGuildProvider } from "components/create-guild/CreateGuildContext"
+import { CreateGuildDynamicDevTool } from "components/create-guild/CreateGuildDynamicDevTool"
+import { CreateGuildHead } from "components/create-guild/CreateGuildHead"
+import { CreateGuildHeadline } from "components/create-guild/CreateGuildHeadline"
+
+const CreateGuildPage = (): JSX.Element => (
+  <>
+    <Layout.Root>
+      <CreateGuildHead />
```
dovalid, line 13: It's really great that the CreateGuildContent and the general layout logic is separated now, but putting each layout component in a separate wrapper component is unnecessary, I find it too much abstraction. We should simply just inline them here, so the page structure (including related logic and props) is easy to see through, just keep CreateGuildContent separately. If the Head and Headline components need to be separate too because of the performance problems, keep them as one separate component too (we can put Head in Headline as NextHead can be rendered anywhere)

dominik-stumpf: > just keep CreateGuildContent separately

Do you mean the other `CreateGuild` components should reside in the `create-guild.tsx`? 

We need to have separate wrapper components for them because they could update each other unnecessarily.

PR #1208, opened 2024-05-13
[GUILD-2228] - Gateables & Discord setup flow optimization

cs-balazs:
- `/gateables` can now return `isGuilded`, and `guildId` params (latter one is set when the former one is `true`)
  - On FE, we no longer make guild requests for each entity, instead use these returned props
  - Added `revalidateIfStale: false` to gateables data
- Gateables types now come from `@guildxyz/types`
- Discord setup flow changes:
  - `DiscordGuildSetup`: Removed the `useForm`, which was used for the state management of the selected server. It now only uses a `useState`, which holds the whole data (`{ id, name, img }`) instead of just the id
  - `DCServerCard` now has an `onSubmit` property. It is called when the permission check and role order checks both pass (as opposed to `onSelect`, which is called when the select button is clicked)
  - The permission, and role order check flags have been moved inside `useServerPermissions`. It turned out cleaner this way
     - This hook now has a baked-in `revalidateOnMount: false`. Which is fine for now, but if we end up using this hook elsewhere it might start top be confusing (for now, we only revalidate this data on user action)
  - Most of the data-fetching logic has been moved inside `ServerSetupCard`

- `RepoCard` & `GoogleDocCard`: Almost the same changes in both, they now use the new gateables props
- `useCountdownSeconds` also had some changes, because it couldn't restart the countdown, and didn't have a proper loading state 😄 

- BE: https://github.com/guildxyz/guild-backend/pull/1454

~~Build is expected to fail until BE is merged & `@guildxyz/types` is bumped~~


BrickheadJohnny (approved):
I really like these changes! That `existingPermissions || (await mutate())` looks a bit strange, but I don't have any alternative idea, so it's good I guess 😉 
Also, we might be able to re-use the `mapPermissions` function inside the `DiscordBotPermissionChecker` component, but that's out of scope for this PR.

PR #1216, opened 2024-05-14
fix(useSetRoleImageAndNameFromPlatformData): add early return to useEffect

BrickheadJohnny:
This hook ran too many times & called the Pinata API constantly, which is not the expected behavior. As a quick fix, I added an extra `alreadyUploaded` state & an early return, and also added a condition so we don't re-upload the image if we already uploaded it to IPFS. Let me know if you have a better idea, not sure how could we handle this properly.

### DIFF
--- a/src/components/[guild]/AddRewardButton/hooks/useSetRoleImageAndNameFromPlatformData.ts
+++ b/src/components/[guild]/AddRewardButton/hooks/useSetRoleImageAndNameFromPlatformData.ts
@@ -3,13 +3,14 @@ import {
   useAddRewardContext,
 } from "components/[guild]/AddRewardContext"
 import usePinata from "hooks/usePinata/usePinata"
-import { useEffect } from "react"
+import { useEffect, useState } from "react"
 import { useFormContext } from "react-hook-form"
 
 const useSetRoleImageAndNameFromPlatformData = (
   platformImage: string,
   platformName: string
 ) => {
+  const [alreadyUploaded, setAlreadyUploaded] = useState(false)
   const { activeTab } = useAddRewardContext()
 
   const { setValue } = useFormContext()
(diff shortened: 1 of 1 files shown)


src/components/[guild]/AddRewardButton/hooks/useSetRoleImageAndNameFromPlatformData.ts
```diff
@@ -3,13 +3,14 @@ import {
...
 import usePinata from "hooks/usePinata/usePinata"
-import { useEffect } from "react"
+import { useEffect, useState } from "react"
 import { useFormContext } from "react-hook-form"
 
 const useSetRoleImageAndNameFromPlatformData = (
   platformImage: string,
   platformName: string
 ) => {
+  const [alreadyUploaded, setAlreadyUploaded] = useState(false)
```
Comment (cs-balazs) [inline src/components/[guild]/AddRewardButton/hooks/useSetRoleImageAndNameFromPlatformData.ts:13]: I'm wondering if this state should be bound to a `platformImage`, `platformName` pair somehow? 🤔 
Not sure, but this way if the same hook call gets new params, we won't upload on the second one? Not sure 😅

cs-balazs (approved):
Nice fix 💯 Left one question regarding the `alreadyUploaded` state, but that might be an irrelevant scenario, if it works for simple ugase, I think we should :shipit: 
I don't have a better idea, I guess the more reliable solution would be to attach this logic to the appropriate `onSuccess` callbacks 🤔

src/components/[guild]/AddRewardButton/hooks/useSetRoleImageAndNameFromPlatformData.ts
Comment (BrickheadJohnny) [reply src/components/[guild]/AddRewardButton/hooks/useSetRoleImageAndNameFromPlatformData.ts:13]: Oh yep, in my opinion that isn't a big problem actually, since the users can always upload a new role image later, but we can consider reworking this hook later. For now I think I'll merge it and add an issue to Linear about this 😉

PR #1218, opened 2024-05-15
refactor(useEditGuildPlatform): send the whole `platformGuildData` object to our API

BrickheadJohnny:
Our API rewrites the `platformGuildData` object when we edit it, so we always need to send both the unmodified and modified data. I refactored this hook, so we won't need to handle this logic where we use the `useEditGuildPlatform` hook.

### DIFF
--- a/src/components/[guild]/AccessHub/hooks/useEditGuildPlatform.ts
+++ b/src/components/[guild]/AccessHub/hooks/useEditGuildPlatform.ts
@@ -1,9 +1,16 @@
 import useGuild from "components/[guild]/hooks/useGuild"
+import useGuildPlatform from "components/[guild]/hooks/useGuildPlatform"
 import useShowErrorToast from "hooks/useShowErrorToast"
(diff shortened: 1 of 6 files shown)


src/components/[guild]/AccessHub/hooks/useEditGuildPlatform.ts
```diff
@@ -13,16 +20,30 @@ const useEditGuildPlatform = ({
...
+    fetcherWithSign([
+      `/v2/guilds/${id}/guild-platforms/${guildPlatformId}`,
+      {
+        method: "PUT",
+        /**
+         * We need to send the whole platformGuildData here, since our API replaces
+         * this object with the data we send inside the request body
+         */
+        body: {
+          platformGuildData: {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/AccessHub/hooks/useEditGuildPlatform.ts:38]: We just send the `platformGuildData` to the API here, but I think that's fine, since we won't need to edit anything else on a `guildPlatform`.

PR #1222, opened 2024-05-18
dynamic reward setup flow from role editor

FBalint:


src/components/[guild]/RolePlatforms/RolePlatforms.tsx
```diff
@@ -33,7 +34,11 @@ type Props = {
 export const openRewardSettingsGuildPlatformIdAtom = atom(0)
 
 const RolePlatforms = ({ roleId }: Props) => {
-  const { onOpen } = useAddRewardContext()
+  const { onOpen, setTargetRoleId } = useAddRewardContext()
+
+  useEffect(() => {
```
Comment (dovalid) [inline src/components/[guild]/RolePlatforms/RolePlatforms.tsx:39]: As far as I can see we don't need a useEffect and a state for this in the provider, we can just pass it as a prop in RolePlatformsWrapper

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/DynamicSetup/BaseValueModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+  return (
+    <>
+      <Modal isOpen={isOpen} onClose={onClose} colorScheme={"dark"}>
+        <ModalOverlay />
+        <ModalContent>
+          <ModalHeader pb={4}>Select base value</ModalHeader>
+          <ModalBody>
+            {dynamicRequirements?.length > 0 ? (
+              <>
+                <Text fontWeight={"semibold"} color={"GrayText"} mb={5}>
```
Comment (dovalid) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/DynamicSetup/BaseValueModal.tsx:73]: It'd be great to give more context here that 1. we need requirements to provide values so they will influence the access to the role, and 2. only specific requirements provide values and are selectable

src/platforms/Points/usePointsCardProps.tsx
```diff
@@ -1,18 +1,22 @@
...
+    : rolePlatform?.platformRoleData?.score
 
   return {
     type: "POINTS" as PlatformName,
     image: imageUrl || <Star />,
     // if undefined at admin setup -> "some", if saved with no value (empty string) -> 0
-    name: `Get ${(rolePlatform?.platformRoleData?.score ?? "some") || 0} ${
-      name || "points"
-    }`,
+    name: `Get ${score ?? 0} ${name || "points"}`,
```
dovalid, line 19: I think we should show the DynamicTag here too similarly as in PointsReward, cause now the card can be misleading for dynamic points (looks like it's static with the shown value set)

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/DynamicSetup/DynamicRewardSetup.tsx
```diff
@@ -0,0 +1,112 @@
...
+  )
+
+  const { onOpen, onClose, isOpen } = useDisclosure()
+
+  return (
+    <>
+      <Stack gap={4}>
+        <Box>
+          <HStack justifyContent={"space-between"} w="full" mb="2">
+            <FormLabel mb="0">Base value</FormLabel>
```
Comment (dovalid) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/DynamicSetup/DynamicRewardSetup.tsx:57]: Would be great to add error handling here, so we show red FormErrorMessage if the user tries to continue without setting a requirement

src/components/[guild]/RolePlatforms/RolePlatforms.tsx
Comment (FBalint) [reply src/components/[guild]/RolePlatforms/RolePlatforms.tsx:39]: I am not sure about that, the targetRoleId is needed in multiple components, should we prop drill to all of them?

src/platforms/Points/usePointsCardProps.tsx
FBalint: The user's points are not necessarily calculated from a single dynamic reward, they are summed from all of the point rewards. The dynamic modal only displays the calculation for a single specific reward, it would be very complicated to display a calculation accross all of them here.

src/components/[guild]/RolePlatforms/RolePlatforms.tsx
Comment (dovalid) [reply src/components/[guild]/RolePlatforms/RolePlatforms.tsx:39]: No, I mean we can pass it to the provider as a prop and just put in into the context there so it's available in every children, don't need the state and the useEffect for it

PR #1226, opened 2024-05-22
fix: avoid snapshot name duplication

FBalint:


cs-balazs (approved):
Nice! I'm not very familiar with this part of the codebase, but the changes make sense! 🚀

PR #1233, opened 2024-05-24
fix: enable switching between dynamic and static on points edit

FBalint:


dovalid (commented):
Seems good, just two things:
- now I always get `Missing linked requirement` warning on the card even tho I've set a requirement, we should look into that
- we could remove the relevant dynamic data part this way from AddPointsPanel and just spread dynamicAmount there since it's already set here, no?

FBalint:
> Seems good, just two things:
> 
> * now I always get `Missing linked requirement` warning on the card even tho I've set a requirement, we should look into that
> * we could remove the relevant dynamic data part this way from AddPointsPanel and just spread dynamicAmount there since it's already set here, no?

This PR was aimed to fix the first issue:
https://github.com/guildxyz/guild.xyz/pull/1232

FBalint:
> > Seems good, just two things:
> > 
> > * now I always get `Missing linked requirement` warning on the card even tho I've set a requirement, we should look into that
> > * we could remove the relevant dynamic data part this way from AddPointsPanel and just spread dynamicAmount there since it's already set here, no?
> 
> This PR was aimed to fix the first issue: #1232

Oh, and the other point - the AddPointsPanel only sets the dynamic part if its needed. When the user sets up a static reward, if we filled the dynamicAmount part nonetheless, it would turn into a dynamic reward.

dovalid:
> This PR was aimed to fix the first issue: #1232

Hmm, I've merged it now but still have the issue

> Oh, and the other point - the AddPointsPanel only sets the dynamic part if its needed. When the user sets up a static reward, if we filled the dynamicAmount part nonetheless, it would turn into a dynamic reward.

Yeah but the switch logic handles that too, no? We only set the dynamicAmount part if we switch to it, and we remove it if we switch back to static

PR #1236, opened 2024-05-27
feat: add dynamic value provider requirements

FBalint:


Mazzika1 (commented):
Okay

Mazzika1 (commented):
Okay

src/requirements/WalletActivity/hooks/ContractDeployProvidedValue.tsx
```diff
@@ -0,0 +1,20 @@
+import { HStack, Text, useColorModeValue } from "@chakra-ui/react"
+import RequirementChainIndicator from "components/[guild]/Requirements/components/RequirementChainIndicator"
+import { ProvidedValueDisplayProps } from "requirements"
+
+const ContractDeployProvidedValue = ({ requirement }: ProvidedValueDisplayProps) => {
```
dovalid, line 5: Why are these components in hooks folder? We should rename the folders to providedValue imo (and make a separate folder when there're real hooks too)

src/components/[guild]/Requirements/components/DataProviderRequirement.tsx
```diff
@@ -50,6 +57,9 @@ const DataProviderRequirement = ({
           <Text
             fontWeight={"medium"}
             sx={{ fontSize: "sm", "& *": { fontSize: "inherit" } }}
+            whiteSpace={"nowrap"}
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/DataProviderRequirement.tsx:60]: These three props could just be `noOfLines={1}`

src/requirements/WalletActivity/hooks/ContractDeployProvidedValue.tsx
FBalint: Originally, these were hooks instead of components, similar to how the rewards have their own hooks. Ive forgotten to update these, thanks!

src/components/[guild]/Requirements/components/DataProviderRequirement.tsx
Comment (FBalint) [reply src/components/[guild]/Requirements/components/DataProviderRequirement.tsx:60]: niiiiiice 🔥

PR #1239, opened 2024-06-03
[GUILD-3074] - List already added DC servers in setup grid

cs-balazs:
This PR adds the functionality for selecting DC servers, which are already added to a guild
- Both adding with the guild page _add reward_ button, and to an existing role should work, if you try the preview out, please check out both
- When adding through the guild page _add reward_ button, only those roles should be listed, which don't already have the selected server as a reward

### DIFF
--- a/src/components/[guild]/AddRewardButton/hooks/useSubmitAddReward.ts
+++ b/src/components/[guild]/AddRewardButton/hooks/useSubmitAddReward.ts
@@ -106,6 +106,8 @@ const useSubmitAddReward = () => {
 
     const existingDCReward = guildPlatforms?.find(
       (gp) =>
+        data?.rolePlatforms?.[0]?.guildPlatform?.platformId ===
+          PlatformType.DISCORD &&
         data?.rolePlatforms?.[0]?.guildPlatform?.platformId === gp.platformId &&
         data?.rolePlatforms?.[0]?.guildPlatform?.platformGuildId ===
           gp.platformGuildId
--- a/src/components/[guild]/AddRewardButton/hooks/useSubmitAddReward.ts
+++ b/src/components/[guild]/AddRewardButton/hooks/useSubmitAddReward.ts
@@ -11,6 +11,7 @@ import { defaultValues } from "../AddRewardButton"
 import useCreateReqBasedTokenReward from "../useCreateTokenReward"
 import useAddReward from "./useAddReward"
 import { useAddRewardDiscardAlert } from "./useAddRewardDiscardAlert"
+import useAddRoleRewards from "./useAddRoleRewards"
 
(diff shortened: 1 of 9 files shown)


src/components/common/DiscordGuildSetup/components/DCServerCard.tsx
```diff
@@ -64,13 +62,13 @@ const DCServerCard = ({
...
-        {isSelected ? (
+        {!isUsedInCurrentGuild && isSelected ? (
           <Button h={10} onClick={onCancel}>
             Cancel
           </Button>
-        ) : isValidating ? (
+        ) : !isUsedInCurrentGuild && isValidating ? (
           <Button isLoading />
-        ) : serverData.isGuilded ? (
+        ) : !isUsedInCurrentGuild && serverData.isGuilded ? (
```
BrickheadJohnny, line 71: Maybe we could refactor this logic, because right now this line means `!(serverData.isGuilded && serverData.guildId === id) && serverData.isGuilded`, which is a bit strange :D What do you think?

BrickheadJohnny (approved):
Left 2 minor suggestions, but LGTM overall!

src/components/[guild]/AddRewardButton/hooks/useSubmitAddReward.ts
```diff
@@ -92,6 +104,23 @@ const useSubmitAddReward = () => {
   const onSubmit = async (data: any, saveAs: "DRAFT" | "PUBLIC" = "PUBLIC") => {
     if (isERC20(data)) return submitERC20Reward(data, saveAs)
 
+    const existingDCReward = guildPlatforms?.find(
+      (gp) =>
+        data?.rolePlatforms?.[0]?.guildPlatform?.platformId === gp.platformId &&
```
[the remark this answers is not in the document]
Comment (cs-balazs) [reply src/components/[guild]/AddRewardButton/hooks/useSubmitAddReward.ts:111]: That is a nice catch, we absolutely need that! Added in https://github.com/guildxyz/guild.xyz/pull/1239/commits/e7665880424b5eeace713876b4539572750ebd9a!

src/components/common/DiscordGuildSetup/components/DCServerCard.tsx
cs-balazs: Agreed, made some changes in https://github.com/guildxyz/guild.xyz/pull/1239/commits/ee6e118a42fb26163662a3d5a1a52399a944495e I will have to test if functionally it is the same, once the preview is deployed
`selectButton` appears twice, but I think this way we have the cleanest conditions, what do you think?

BrickheadJohnny: Yup, this looks much cleaner to me! Thanks!

PR #1242, opened 2024-06-03
refactor: useScrollBatchedRendering hook

### LINKED ISSUE #629: refactor: useScrollBatchedRendering hook
We use scroll based lazy rendering in 3 places now, should abstract the relevant logic into a common hook. Search for `// TODO: we use this behaviour in multiple places now, should make a useScrollBatchedRendering hook`

We should implement it with a global state so we can get back to where we were at on page navigation (as it's done for the explorer right now with context)

dominik-stumpf:
Closes #629


src/hooks/useScrollBatchedRendering.ts
```diff
@@ -0,0 +1,19 @@
+import { Dispatch, MutableRefObject, SetStateAction } from "react"
+import useScrollEffect from "./useScrollEffect"
+
+export function useScrollBatchedRendering(
```
BrickheadJohnny, line 4: I have a couple of ideas here:
- Maybe we could move the `elementCount` state inside this hook, and return its value, that way we wouldn't need to handle the state everywhere where we use this new hook
- We could try to create a ref and return it from the hook. I'd keep the `scrollTarget` property (if it's defined, use that one, if it isn't defined, use the "locally" created ref) too, because we need to pass down the ref to this hook inside the `ExploreAllGuilds` component (boy scout: type the `ref` inside that file 😉)
- I'd also consider using an object as the param of this hook, so we could use it this way (it'd be easier to understand/read imo):

```ts
const { ref, elementCount } = useScrollBatchedRendering({
  batchSize: 20,
  disableRendering: !someVariable
})
```

What do you think?

BrickheadJohnny (commented):
Left a couple of suggestions, let me know what do you think about them!

src/hooks/useScrollBatchedRendering.ts
dominik-stumpf: > - Maybe we could move the `elementCount` state inside this hook, and return its value, that way we wouldn't need to handle the state everywhere where we use this new hook

In my first attempt I tried to do that but the `disableRendering` is not same everywhere so I would have to move that logic inside the hook along with the state, and sync external states outside if there is any. I mean here the case of `setSize` from `ExploreAllGuilds.tsx` specifically.

Consequently this abstraction would demand each `disableRendering` case to be defined inside the hook, which I find hard to extend upon later.

BrickheadJohnny: Oh, I see, that `setSize` comes from the `useSWRInfinite` call. And what if we still move the `elementsCount` state inside the hook, but also add an `onElementsCountChange` parameter and call it when we update the state? That way we'd have a nice default (managing the state inside the hook & returning it), but we can also update some external state optionally. We shouldn't do anything with the `disableRendering` variable, we can provide it to the hook just like we do it now. Wdyt?

dominik-stumpf: On `ExploreAllGuilds.tsx` the `onElementsCountChange` idea would work, however I can't have the state in, while having the condition out, since in other cases it relies on the element. Here is an example:

 ```tsx
 // const [elementCount, setRenderedMembersCount] = useState(BATCH_SIZE)
  const disableRendering = useMemo(
    () => members?.length <= elementCount,
    [members, elementCount]
  )
  const {scrollTarget, elementCount} = useScrollBatchedRendering({
    batchSize: BATCH_SIZE,
    disableRendering,
    // setElementCount: setRenderedMembersCount,
  })
  ```
  
  The only solution I have in mind is when I move in the `members` variable in and have the hook handle conditions. But again, that makes it hard to extend.

BrickheadJohnny: Oh, I see. @dovalid do you have any ideas for this? Or should we just keep it simple and do it as Dominik suggested? (Keep the state outside of the hook)

dovalid: Yep, I agree that we should keep the state outside of the hook this way. Actually, maybe it'd be better to keep the whole setter logic outside, instead of passing batchSize, disableRendering and setElementCount props, just passing a constructed callback containing the condition too, but it's fine this way too so I'd leave it like this for now

src/pages/[guild]/leaderboard/[pointsId].tsx
```diff
@@ -57,16 +56,15 @@ const Leaderboard = () => {
...
-    if (
-      !wrapperRef.current ||
-      wrapperRef.current.getBoundingClientRect().bottom > window.innerHeight ||
-      data?.leaderboard?.length <= renderedUsersCount
-    )
-      return
-
-    setRenderedUsersCount((prevValue) => prevValue + BATCH_SIZE)
-  }, [data, renderedUsersCount])
+  const disableRendering = useMemo(
```
Comment (dovalid) [inline src/pages/[guild]/leaderboard/[pointsId].tsx:59]: Do we need these useMemos? If not, it'd be much nicer to just inline the condition when passing the prop. If we do, consider the PR approved by me

BrickheadJohnny (approved):
🚀

PR #1245, opened 2024-06-05
[GUILD-3132] fix: convert only relevant elements of the url to lowercase

dominik-stumpf:
The `URL` API by default converts elements in the *origin* to lowercase.

### DIFF
--- a/src/requirements/VisitLink/VisitLinkForm.tsx
+++ b/src/requirements/VisitLink/VisitLinkForm.tsx
@@ -57,7 +57,12 @@ const VisitLinkForm = ({ baseFieldPath }: RequirementFormProps) => {
         placeholder="https://guild.xyz"
         onChange={(e) => {
           const position = e.target.selectionStart
-          onChange(new URL(e.target.value).href)
+          try {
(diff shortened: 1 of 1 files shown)


src/requirements/VisitLink/VisitLinkForm.tsx
```diff
@@ -57,11 +57,7 @@ const VisitLinkForm = ({ baseFieldPath }: RequirementFormProps) => {
         placeholder="https://guild.xyz"
         onChange={(e) => {
           const position = e.target.selectionStart
-          onChange(
-            e.target.value.toLowerCase().includes("youtube")
-              ? e.target.value
-              : e.target.value.toLowerCase()
-          )
+          onChange(new URL(e.target.value).href)
```
BrickheadJohnny, line 60: Lovely solution, but this will cause an error:
image

Maybe you could just put it in a try-catch like this?

```ts
try {
  const newValue = new URL(e.target.value).href
  onChange(newValue)
} catch {
  onChange(e.target.value)
}
```

BrickheadJohnny (approved):

PR #1246, opened 2024-06-05
Add nvmrc

PumpkinSeed:


BrickheadJohnny (approved):
Very nice programming.

Do you think we should also add preinstall/prebuild scripts (e.g. `"preinstall": "nvm use"` to package.json, or we should just manually run `nvm use` on demand?

PumpkinSeed:
> Do you think we should also add preinstall/prebuild scripts (e.g. `"preinstall": "nvm use"` to package.json, or we should just manually run `nvm use` on demand?

I think that is enough if we run it manually for now.

dovalid (approved):

PR #1247, opened 2024-06-05
Fix dynamic reward edit submit bug

FBalint:


FBalint:
The solution was found on the forum, so Im taking the liberty of approving and merging my own PR

PR #1248, opened 2024-06-05
[GUILD-3104] - WaaS: Change backup data export feature to private key export

cs-balazs:
[Linear](https://linear.app/guildxyz/issue/GUILD-3104/coinbase-data-import) - [Discord](https://discord.com/channels/697041998728659035/1169587223956570232/1247473523816140821)

- This is needed for the user linked in the Linear ticket, and once we transition to Smart Wallet, we will need this option for current waas users
- Can be tested by:
  1. Connecting with waas
  2. Taking note of the address
  3. Exporting the private key
  4. Importing the wallet into MetaMask (or other wallet)
  5. Checking if the imported wallet's address is the same as the one noted earlier


BrickheadJohnny (approved):
Couldn't test it easily on Vercel (Google auth redirect URL mismatch), but the code itself looks good to me, and according to the PR description, you already tested it, so LFG 🚀

PR #1249, opened 2024-06-05
[GUILD-2197] - refactor: move complicated `useEffect` logic in `useUserPublic`

cs-balazs:
- This seems to be a much simpler solution. One less `useEffect` always feels like a good idea
- Tested by:
  - I connected my wallet, refreshed the page, and the modal appeared
  - I invalidated my key, and reloaded the page, and both the toast appears, and the modal opened
  - Temporarily added `/explorer` to the ignored routes array, and the modal no longer open

### DIFF
--- a/src/components/[guild]/hooks/useUser.ts
+++ b/src/components/[guild]/hooks/useUser.ts
@@ -8,10 +8,20 @@ import {
 } from "hooks/useSetKeyPair"
 import useToast from "hooks/useToast"
 import { useSetAtom } from "jotai"
+import { useRouter } from "next/router"
 import { KeyedMutator } from "swr"
 import useSWRImmutable from "swr/immutable"
 import { User } from "types"
 import fetcher, { useFetcherWithSign } from "utils/fetcher"
+
+// We don't open WaletSelectorModal on these routes
+const ignoredRoutes = [
+  "/_error",
+  "/tgauth",
+  "/oauth",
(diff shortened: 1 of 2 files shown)


src/components/[guild]/hooks/useUser.ts
```diff
@@ -90,11 +101,16 @@ const useUserPublic = (
...
           })
-
-          setIsWalletSelectorModalOpen(true)
         }
       }
 
+      // If we didn't set the keyPair field, the user either doesn't have one locally, or has an invalid one
+      const shouldOpenWalletSelectorModal = !user.keyPair
+
+      if (shouldOpenWalletSelectorModal && !ignoredRoutes.includes(router.route)) {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/hooks/useUser.ts:110]: nit, but I think the comment explains this logic pretty well, so we don't need an extra variable here, we can just check `if(!user.keyPair)`, wdyt?

BrickheadJohnny (approved):
Beautiful! 🚀

src/components/[guild]/hooks/useUser.ts
Comment (cs-balazs) [reply src/components/[guild]/hooks/useUser.ts]: Agreed, the comment should be enough, thanks for pointing it out!
Changed in https://github.com/guildxyz/guild.xyz/pull/1249/commits/5e1a3d66d989996957574e36a710ce330918d1a0

PR #1250, opened 2024-06-05
feat: set role name from reward setup

FBalint:


src/components/[guild]/AddRewardButton/AddRewardButton.tsx
```diff
@@ -21,6 +21,7 @@ export type AddRewardForm = {
   requirements?: Requirement[]
   roleIds?: number[]
   visibility: Visibility
+  name?: string // Name for role, if new role is created with reward
```
Comment (BrickheadJohnny) [inline src/components/[guild]/AddRewardButton/AddRewardButton.tsx:24]: We could call this `roleName` so we wouldn't need to explain it in a comment, what do you think?

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddFormPanel/AddFormPanel.tsx
```diff
@@ -29,6 +29,7 @@ const AddFormPanel = ({ onAdd }: AddRewardPanelProps) => {
   useAddRewardDiscardAlert(methods.formState.isDirty)
 
   const fields = useWatch({ control: methods.control, name: "fields" })
+  const name = useWatch({ control: methods.control, name: "name" })
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddFormPanel/AddFormPanel.tsx:32]: We could also use `roleName` here, that won't be confusing at all imo ;)

BrickheadJohnny (approved):
LGTM overall, only left 2 minor suggestions!

PR #1256, opened 2024-06-06
monitor add reward button usage

FBalint:
To understand how people utilize this feature, I configured PostHog events to record the settings applied when they add rewards using the button.


BrickheadJohnny (approved):
🚀

PR #1257, opened 2024-06-06
[GUILD-2958] prevent multiple addRequirement calls

dominik-stumpf:


src/requirements/Payment/PaymentForm.tsx
```diff
@@ -62,9 +62,13 @@ const PaymentForm = ({
       setValue(`${baseFieldPath}.data.id`, registeredVaultId),
   })
 
+  const isRequirementAdded = useRef(false)
   useEffect(() => {
     if (!vaultId) return
-    addRequirement()
+    if (!isRequirementAdded.current) {
```
BrickheadJohnny, line 68: I wonder if we could completely get rid of this `useEffect`? Maybe if we set `${baseFieldPath}.data.id` in `useRegisterVault`'s `onSuccess`, and call `addRequirement` there too? Not sure if that'd work properly, but I think we could try something like that, it'd simplify the code a lot. 😉

src/requirements/Payment/PaymentForm.tsx
```diff
@@ -58,19 +58,14 @@ const PaymentForm = ({
     token,
     fee,
     owner,
-    onSuccess: (registeredVaultId) =>
-      setValue(`${baseFieldPath}.data.id`, registeredVaultId),
+    onSuccess: (registeredVaultId) => {
+      setValue(`${baseFieldPath}.data.id`, registeredVaultId)
+      if (registeredVaultId) {
```
BrickheadJohnny, line 63: I believe `registeredVaultId` will always be defined here (see the `useRegisterVault` hook for more details), so it's safe to remove this check.

BrickheadJohnny (approved):
Left one minor comment, but seems like it works this way, so approving it, thanks! 🚀

PR #1258, opened 2024-06-06
Improve and fix `PaymentForm`, prevent JS scientific notation conversion

dominik-stumpf:
This PR prevents error: https://jam.dev/c/d88f9e28-bddf-46f4-a8a7-5a668a43b3c8

Added validation to make sure the token precision is not exceeded, `parseFloat` is replaced by string as JS numbers are unreliable for this kind of use case.

### DIFF
--- a/src/requirements/Payment/PaymentForm.tsx
+++ b/src/requirements/Payment/PaymentForm.tsx
@@ -2,7 +2,7 @@ import { HStack, Icon, Stack, Tooltip } from "@chakra-ui/react"
 import Button from "components/common/Button"
 import useTriggerNetworkChange from "hooks/useTriggerNetworkChange"
 import { Check, Question } from "phosphor-react"
-import { useEffect } from "react"
+import { useEffect, useRef } from "react"
 import { FormProvider, useForm, useFormContext, useWatch } from "react-hook-form"
 import { RequirementFormProps } from "requirements"
 import { FEE_COLLECTOR_CONTRACT } from "utils/guildCheckout/constants"
@@ -62,9 +62,13 @@ const PaymentForm = ({
       setValue(`${baseFieldPath}.data.id`, registeredVaultId),
   })
 
+  const isRequirementAdded = useRef(false)
   useEffect(() => {
     if (!vaultId) return
-    addRequirement()
+    if (!isRequirementAdded.current) {
+      addRequirement()
+      isRequirementAdded.current = true
+    }
   }, [vaultId, addRequirement])
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


src/requirements/Payment/PaymentForm.tsx
```diff
@@ -62,9 +62,13 @@ const PaymentForm = ({
       setValue(`${baseFieldPath}.data.id`, registeredVaultId),
   })
 
+  const isRequirementAdded = useRef(false)
```
BrickheadJohnny, line 65: Whoopsie, this got here from another PR hehe

src/requirements/Payment/components/RegisterVaultForm/RegisterVaultForm.tsx
```diff
@@ -49,6 +51,29 @@ const RegisterVaultForm = ({ isDisabled }: Props): JSX.Element => {
     formState: { errors },
   } = useFormContext<RegisterVaultFormType>()
   const chain = useWatch({ name: "chain" })
+  const token = useWatch({ name: "token" })
+
+  const { data: tokenData } = useToken({
+    address: token,
+    chainId: Chains[chain as keyof typeof Chains],
```
BrickheadJohnny, line 58: [We have a `Chain` type for this](https://github.com/guildxyz/guild.xyz/blob/main/src/wagmiConfig/chains.ts#L698) ;)

But I see the root of the problem here & we probably don't need a typecast. We can either add generics to `useWatch` like this:

```ts
const chain = useWatch<RegisterVaultFormType, "chain">({name: "chain"})
```

Or just pass `control` from `useFormContext` to `useWatch`:

```ts
const {
    control,
    register,
    setValue,
    formState: { errors },
  } = useFormContext<RegisterVaultFormType>()
  
const chain = useWatch({ control, name: "token" })
```

I believe either of these solutions will work, so use whichever one you prefer!

src/requirements/Payment/components/RegisterVaultForm/RegisterVaultForm.tsx
```diff
@@ -98,13 +125,14 @@ const RegisterVaultForm = ({ isDisabled }: Props): JSX.Element => {
 
         <InputGroup>
           <NumberInput
-            isDisabled={isDisabled}
+            isDisabled={isDisabled || !token || !chain}
```
BrickheadJohnny, line 128: I'd maybe disable this input until `tokenData` is not defined, what do you think?

BrickheadJohnny (commented):
Great work overall, left some improvement ideas 😉

BrickheadJohnny (approved):
Great work!

BrickheadJohnny:
@dominik-stumpf I think we can merge this one too! 🚀

PR #1261, opened 2024-06-07
feat: interactive `TiltCard` component

BrickheadJohnny:
✨


dovalid (approved):
Beautiful solution, I love this

dovalid (commented):
Only thing I've noticed is at the setup flow if I click on the image, the glare disappears, until I click somewhere else. Idk if it's expected because of something / is easily fixable, we should only fix it if it's trivial, otherways it's not important at all

BrickheadJohnny:
> Only thing I've noticed is at the setup flow if I click on the image, the glare disappears, until I click somewhere else. Idk if it's expected because of something / is easily fixable, we should only fix it if it's trivial, otherways it's not important at all

Oh yup, for some reason, a `z-index: 1` prop is added to the Button element in focused state, that's why `mix-blend-mode` doesn't work in that case. I think I'll merge the PR as is, it's pretty unpredictable if a child component will be in the same stacking context or not. :S Thanks for the review!

PR #1264, opened 2024-06-07
[GUILD-3156] feat: add team and sdk to NavMenu

dominik-stumpf:


BrickheadJohnny (commented):
I wonder if we should move the "Guide" link to the 1st column to make the layout more balanced? What do you think?
image

dominik-stumpf:
> I wonder if we should move the "Guide" link to the 1st column to make the layout more balanced? What do you think? image

Certainly! I thought *Navigation* contained all the "local domain" redirections, but *Guide* seems to fit there nicely.

BrickheadJohnny (approved):

PR #1265, opened 2024-06-10
feat: add middleware to lowercase the request URL

BrickheadJohnny:


FBalint:
Looks great!

One nit: maybe we should decode the URI before setting it to lowercase. This way, percent encoded characters would also get lowercased.

cs-balazs (approved):
Awesome! 💯

BrickheadJohnny:
> Looks great!
> 
> One nit: maybe we should decode the URI before setting it to lowercase. This way, percent encoded characters would also get lowercased.

Good point, updated the logic in https://github.com/guildxyz/guild.xyz/pull/1265/commits/d532ff181d080a8d4e4ae47c4d692abe49ee1fa4

PR #1271, opened 2024-06-12
fix(WalletSelectorModal): show toast if user tries to link their current address

BrickheadJohnny:
Users sometimes try to link the address which they currently use to their account - this makes no sense & this was an unhandled path (we opened the link address modal without a "Link address" button). This PR aims to solve this by showing an info toast with some guidance & hiding the Link Address modal in this case.


cs-balazs (approved):
🔥

PR #1274, opened 2024-06-12
Address `strictNullChecks` errors on `src/utils`

dominik-stumpf:
This PR fixes 37 errors.

### DIFF
--- a/src/utils/fetcher.ts
+++ b/src/utils/fetcher.ts
@@ -73,7 +73,11 @@ const fetcher = async (
         location?.reload()
       }
 
-      if (isGuildApiCall || resource.includes(process.env.NEXT_PUBLIC_API)) {
+      if (
+        isGuildApiCall ||
+        (process.env.NEXT_PUBLIC_API &&
+          resource.includes(process.env.NEXT_PUBLIC_API))
+      ) {
         const error = res.errors?.[0]
 
         const errorMsg = error
--- a/src/utils/mapRequirement.ts
+++ b/src/utils/mapRequirement.ts
@@ -2,12 +2,17 @@ import { Requirement } from "types"
 
 const mapRequirement = (requirement?: Requirement) => {
   // Using structuredClone so we don't modify the original requirement unintentionally
-  const newRequirement: Requirement = structuredClone(requirement)
+  const newRequirement = structuredClone(requirement) as Partial<Requirement>
+  if (!requirement || !newRequirement) return
 
   if (requirement.type === "COIN")
     newRequirement.address = "0x0000000000000000000000000000000000000000"
 
-  if (newRequirement.type === "CONTRACT" && Array.isArray(requirement.data.params)) {
+  if (
+    newRequirement.type === "CONTRACT" &&
+    Array.isArray(requirement?.data?.params) &&
+    newRequirement.data
+  ) {
     newRequirement.data.params = requirement.data.params.map((param) =>
       typeof param === "string"
         ? {
--- a/src/utils/mapRequirement.ts
+++ b/src/utils/mapRequirement.ts
@@ -2,7 +2,7 @@ import { Requirement } from "types"
 
 const mapRequirement = (requirement?: Requirement) => {
   // Using structuredClone so we don't modify the original requirement unintentionally
-  const newRequirement = structuredClone(requirement) as Partial<Requirement>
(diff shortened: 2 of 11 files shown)


src/utils/preprocessRequirement.ts
```diff
@@ -38,18 +38,24 @@ const preprocessRequirement = (requirement: Partial<Requirement>): Requirement =
   }
 
   if (processedRequirement.type?.startsWith("COVALENT_")) {
-    if (!processedRequirement?.data?.timestamps?.minAmount) {
+    if (
+      processedRequirement.data &&
+      !processedRequirement?.data?.timestamps?.minAmount
```
BrickheadJohnny, line 43: Since we already check if `processedRequirement.data` is defined, I think we don't need optional chaining here, right? So I'd write something like this: `processedRequirement.data.timestamps?.minAmount`. Same goes for the if statement in line 48.

src/utils/preprocessRequirement.ts
```diff
@@ -58,19 +64,22 @@ const preprocessRequirement = (requirement: Partial<Requirement>): Requirement =
     (processedRequirement.type === "ERC721" ||
       processedRequirement.type === "ERC1155" ||
       processedRequirement.type === "NOUNS") &&
-    !processedRequirement.data.attributes?.length &&
+    !processedRequirement?.data?.attributes?.length &&
     !processedRequirement.data?.ids?.length
   ) {
-    processedRequirement.data.attributes = undefined
-    if (!requirement.data.minAmount) processedRequirement.data.minAmount = 0
+    if (processedRequirement.data) {
```
BrickheadJohnny, line 70: We could put this condition in the if statement above.

src/utils/preprocessRequirement.ts
```diff
@@ -58,19 +64,22 @@ const preprocessRequirement = (requirement: Partial<Requirement>): Requirement =
...
+    }
   }
 
   if (
     (processedRequirement.type === "ERC721" ||
       processedRequirement.type === "ERC1155") &&
     processedRequirement.data?.attributes?.length
   ) {
-    processedRequirement.data.attributes = requirement.data.attributes.map(
+    processedRequirement.data.attributes = requirement?.data?.attributes.map(
```
BrickheadJohnny, line 82: This looks a bit strange, since we check if `processedRequirement.data.attributes.length` is greater than 0, so this array should be defined at this point.

src/utils/preprocessRequirement.ts
```diff
@@ -103,7 +112,7 @@ const preprocessRequirement = (requirement: Partial<Requirement>): Requirement =
     processedRequirement.type === "CONTRACT" &&
     Array.isArray(processedRequirement.data?.params)
   ) {
-    processedRequirement.data.params = requirement.data.params.map(
+    processedRequirement.data.params = requirement?.data?.params.map(
```
BrickheadJohnny, line 115: Same here, it's a bit odd that we check if `requirement.data.params` is an array, but we still need optional chaining? 👀

src/utils/fetcher.ts
```diff
@@ -73,7 +73,11 @@ const fetcher = async (
         location?.reload()
       }
 
-      if (isGuildApiCall || resource.includes(process.env.NEXT_PUBLIC_API)) {
+      if (
+        isGuildApiCall ||
+        (process.env.NEXT_PUBLIC_API &&
```
BrickheadJohnny, line 78: It would be great if we could just use these env vars without checking if they're defined or not. Maybe we could write an env parser which would throw an error if any of the required environment variables is undefined?

src/utils/mapRequirement.ts
```diff
@@ -29,8 +34,6 @@ const mapRequirement = (requirement?: Requirement) => {
   // Removing attributes which we don't need inside the form
   delete newRequirement.createdAt
   delete newRequirement.updatedAt
-  delete newRequirement.symbol
```
BrickheadJohnny, line 32: Hmm, not sure if we can remove these lines, since we actually don't need these props on the client side, but we'll get them from the API for some requirements (e.g. NFT & Token requirements)

dominik-stumpf:
These changes are unstable and `strictNullChecks` is turned back on, so there is no need to do per folder fixes, closing.

PR #1276, opened 2024-06-13
fix(Collect page): always add the relevant guildPlatform to the SWR cache

BrickheadJohnny:
We had a client-side error in cases when the NFT reward wasn't public in a guild. I think we can safely "leak" the `guildPlatform` here, since NFTs are public by default (they're on chain rewards), but let me know if you have any other suggestions!

### DIFF
--- a/src/pages/[guild]/collect/[chain]/[address].tsx
+++ b/src/pages/[guild]/collect/[chain]/[address].tsx
@@ -278,6 +278,10 @@ const getStaticProps = async ({ params }) => {
     .then((res) => res.json())
     .catch(() => undefined)
 
+  const isPublicNFT = !!publicGuild.guildPlatforms.find(
+    (gp) => gp.id === nftGuildReward.id
+  )
+
   return {
     revalidate: 600, // Revalidate at most once every 10 minutes
     props: {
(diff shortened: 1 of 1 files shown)


src/pages/[guild]/collect/[chain]/[address].tsx
```diff
@@ -290,7 +290,11 @@ const getStaticProps = async ({ params }) => {
       fallbackImage: nftGuildReward.platformGuildData.imageUrl,
       // Pre-populating the public guild & requirements caches
       fallback: {
-        [guildPageEndpoint]: publicGuild,
+        [guildPageEndpoint]: {
+          ...publicGuild,
+          // An NFT is always public (it can be found on the blockchain), so we can add the guildPlatform to the public guild cache
+          guildPlatforms: [...publicGuild.guildPlatforms, nftGuildReward],
```
Comment (cs-balazs) [inline src/pages/[guild]/collect/[chain]/[address].tsx:296]: Won't this duplicate it, in the case, when `nftGuildReward` is already public?

Comment (BrickheadJohnny) [reply src/pages/[guild]/collect/[chain]/[address].tsx]: Uh-oh, good point, added some extra logic in https://github.com/guildxyz/guild.xyz/pull/1276/commits/1f9d89080fb1d7f012a2061ac7053bc7932c94fd, let me know if it looks good to you!

Comment (cs-balazs) [reply src/pages/[guild]/collect/[chain]/[address].tsx]: Thanks for the change! ✨

cs-balazs (approved):
Fantastic! 🚀

PR #1277, opened 2024-06-13
Set `strictNullChecks` to true and ignore related errors

dominik-stumpf:


BrickheadJohnny:
We'll go with a simpler solution which won't require these comments, so closing it for now.

PR #1280, opened 2024-06-14
Identify PostHog users with Guild user IDs

BrickheadJohnny:


cs-balazs (approved):
Screenshot 2024-06-14 at 09 21 23

PR #1281, opened 2024-06-14
feat: autofill href if possible and validate

dominik-stumpf:
The validation trigger is controlled in higher up the component tree and it is set to `onChange` which might be unpleasant.

### DIFF
--- a/src/components/[guild]/EditGuild/components/SocialLinks.tsx
+++ b/src/components/[guild]/EditGuild/components/SocialLinks.tsx
@@ -14,8 +14,22 @@ import FormErrorMessage from "components/common/FormErrorMessage"
 import StyledSelect from "components/common/StyledSelect"
 import { Plus } from "phosphor-react"
 import { useFormContext, useWatch } from "react-hook-form"
-import { GuildFormType, SelectOption, supportedSocialLinks } from "types"
+import {
+  GuildFormType,
+  SelectOption,
+  SocialLinkKey,
+  supportedSocialLinks,
+} from "types"
 import capitalize from "utils/capitalize"
+import { z } from "zod"
+
+const socialLinkUserPaths = {
+  TWITTER: "https://twitter.com/",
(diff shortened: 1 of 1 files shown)


src/components/[guild]/EditGuild/components/SocialLinks.tsx
```diff
@@ -14,8 +14,22 @@ import FormErrorMessage from "components/common/FormErrorMessage"
...
+  GuildFormType,
+  SelectOption,
+  SocialLinkKey,
+  supportedSocialLinks,
+} from "types"
 import capitalize from "utils/capitalize"
+import { z } from "zod"
+
+const socialLinkUserPaths = {
+  TWITTER: "https://twitter.com/",
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/components/SocialLinks.tsx:27]: We can use `x.com` here ;)

src/components/[guild]/EditGuild/components/SocialLinks.tsx
```diff
@@ -33,46 +47,69 @@ const SocialLinks = (): JSX.Element => {
...
+            <GridItem key={key}>
+              <FormControl isInvalid={!!errors?.socialLinks?.[key]} isRequired>
+                <InputGroup size="lg">
+                  <InputLeftElement>
+                    {socialLinkOptions.find((sl) => sl.value === key).img}
+                  </InputLeftElement>
+                  <Input
+                    type="url"
+                    {...registerBindings}
+                    onBlur={(...args) => {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/components/SocialLinks.tsx:73]: It's a bit strange that we first validate the field, then we actually change its value on blur, and it becomes valid again. Could we either change it's value in the `onChange` handler or only when the user submits the form? What do you think?

Comment (dominik-stumpf) [reply src/components/[guild]/EditGuild/components/SocialLinks.tsx]: I changed it now so the auto-fill only triggers when new social link is added. I find this approach much more gentle on the user.

BrickheadJohnny (approved):
I like this new approach, thx!

PR #1284, opened 2024-06-17
refactor(AddRoleDrawer): handle dynamic rewards in state

FBalint:
- Until now, dynamic rewards were disabled in the Add Role flow
- This PR aims to solve this by using `RequirementHandlerProvider`s
    - `ClientStateRequirementHandlerProvider`: requirement addition happens in the role form state
    - `ApiRequirementHandlerProvider`: requirement addition happens by calling the API
- Temporary IDs are assigned to the created requirements in the form state (using `Date.now()`)
   - This way, requirements can be referenced by this ID during dynamic reward setup


src/components/[guild]/RoleCard/components/Reward.tsx
```diff
@@ -260,7 +261,11 @@ const RewardWrapper = ({ platform, ...props }: RewardProps) => {
   const Component =
     rewards[PlatformType[guildPlatform?.platformId]].RoleCardComponent ?? Reward
 
-  return <Component platform={platformWithGuildPlatform} {...props} />
+  return (
+    <ApiRequirementHandlerProvider roleId={platformWithGuildPlatform.roleId}>
```
Comment (dovalid) [inline src/components/[guild]/RoleCard/components/Reward.tsx:265]: If I understand correctly this is needed because of the DynamicTag changes, so it works when rendered on the RoleCards too. But this way it's included for every end user too which could negatively affect bundle size & performance, can we use a condition in DynamicReward or a prop based solution instead?

Comment (FBalint) [reply src/components/[guild]/RoleCard/components/Reward.tsx:265]: The problem is that besides the create handler, it also provides the list of requirements itself as well, which is not admin-only.
We can, however, wrap only the `DynamicTag` with the provider, so it would only affect dynamic rewards.

dovalid (approved):

PR #1288, opened 2024-06-18
fix(useGuild): mutate the correct cache key

BrickheadJohnny:
We always mutate the authenticated cache key in `useGuild`'s `onSuccess`, which is not correct, because if a user visits a guild page by the guild's ID (e.g. [guild.xyz/1985](https://guild.xyz/1985)), we just populate the authenticated cache with the unauthenticated data & don't call the `/guilds/:guildId` endpoint with auth, so guild admins can't see the secret/hidden requirements/roles/rewards.

### DIFF
--- a/src/components/[guild]/hooks/useGuild.ts
+++ b/src/components/[guild]/hooks/useGuild.ts
@@ -2,7 +2,7 @@ import useSWRWithOptionalAuth, {
   mutateOptionalAuthSWRKey,
 } from "hooks/useSWRWithOptionalAuth"
 import { useRouter } from "next/router"
-import { unstable_serialize, useSWRConfig } from "swr"
+import { mutate as swrMutate, unstable_serialize, useSWRConfig } from "swr"
 import useSWRImmutable from "swr/immutable"
 import { Guild, SimpleGuild } from "types"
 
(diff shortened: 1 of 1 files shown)


cs-balazs (commented):
I'm wondering why we need any mutate logic in onSuccess at all 🤔 `useSWRWithOptionalAuth` should handle mutations to some degree or? Maybe that hooks needs a revisit soon 😅

BrickheadJohnny:
> I'm wondering why we need any mutate logic in onSuccess at all 🤔 `useSWRWithOptionalAuth` should handle mutations to some degree or? Maybe that hooks needs a revisit soon 😅

We actually need that mutation in this hook only for the Guild requirement types. The idea is that if I visit e.g. the `/our-guild` page, then I visit `johnnys-guild`, which has a requirement to be a member of Our Guild (ID: `1985`), then we should be able to just get `/our-guild`s data from the SWR cache **by ID**, even if we fetched it originally by URL name 😉 

So this is why we had this mutation inside `onSuccess`, but it turns out we didn't handle the case when a user visits a guild page by ID - we simply populated the authenticated cache with the unauthenticated data :hidepainharold:

Let me know if this makes sense now, hehe!

cs-balazs (approved):
That makes sense, thank you! I think my confusion came from thinking that `mutateOptionalAuthSWRKey` handles both authenticated and unauthenticated mutations for some reason 😄

PR #1289, opened 2024-06-18
refactor: add reward flow

FBalint:


src/platforms/components/useSubmitEverything.tsx
```diff
@@ -0,0 +1,220 @@
...
+      showErrorToast("Failed to create role")
+      return null
+    }
+  }
+
+  const createRequirements = async (
+    requirements: Partial<Requirement>[],
+    roleIds: number[]
+  ) => {
+    const requirementIdMap = {}
```
dovalid, line 70: We should add a comment about the two dimensional object, because we can add the same requirement to different roles in the add reward flow, which means they will have different ids per role

FBalint:
The big question - how should we name the hook?
How about `useCreateRRR`? 
As in _Requirement, Role, Reward_.

src/components/[guild]/RoleCard/components/DynamicReward/DynamicTag.tsx
```diff
@@ -12,7 +12,11 @@ const DynamicTag = ({
   ...rest
 }: { rolePlatform: RolePlatform; editDisabled?: boolean } & Rest) => {
   const { isAdmin } = useGuildPermission()
-  const { requirements = null } = useRequirementHandlerContext() || {}
+  const { requirements: contextRequirements } = useRequirementHandlerContext() || {}
+  const { data } = useRequirements(rolePlatform.roleId)
+  const requirements = contextRequirements ?? data
+
+  console.log(rolePlatform.roleId)
```
Comment (dovalid) [inline src/components/[guild]/RoleCard/components/DynamicReward/DynamicTag.tsx:19]: Note: we should remove this before merging

PR #1290, opened 2024-06-18
Set `strictNullChecks` to true by post- and pre-commit hook (alternative attempt)

dominik-stumpf:


BrickheadJohnny:
We'll go with this solution instead: https://github.com/guildxyz/guild.xyz/pull/1291

PR #1293, opened 2024-06-19
I3183 add taiko klaytn fantom

marclllaks:
Adds:
- Taiko with wallet activity
- Klaytn
- Fantom wallet activity


public/networkLogos/taiko.svg
BrickheadJohnny, line 1: We already have a `taiko-katla.svg`, it's the same image, just with white background, could we use that instead?

src/hooks/useTokens.ts
```diff
@@ -74,6 +74,8 @@ export const TokenApiURLs: Record<Chain, string[]> = {
   OP_BNB: [],
   FORM_TESTNET: [],
   CYBER: [],
+  TAIKO: [],
+  KLAYTN: ["https://tokens.coingecko.com/klaytn/all.json"],
```
BrickheadJohnny, line 78: I get an `AccessDenied` error for this 👀

public/networkLogos/taiko.svg
marclllaks: Sure thing, nice catch

src/hooks/useTokens.ts
marclllaks: Yep, this is wrong indeed. We don't have a token list then

BrickheadJohnny (approved):
lfg

PR #1296, opened 2024-06-19
fix(FeesTable): make table wrap words

dominik-stumpf:


src/components/[guild]/Requirements/components/GuildCheckout/components/FeesTable.tsx
```diff
@@ -48,7 +48,13 @@ const FeesTable = ({
       </Button>
 
       <Collapse in={isOpen} animateOpacity>
-        <TableContainer borderWidth={1} borderRadius="xl" mt={2} bgColor={bgColor}>
+        <TableContainer
+          borderWidth={1}
+          borderRadius="xl"
+          mt={2}
+          bgColor={bgColor}
+          whiteSpace={"normal"}
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/components/GuildCheckout/components/FeesTable.tsx:56]: Love this solution, but now the "multiplier" line looks strange (the "x" and the number after it should be in one line)
image

BrickheadJohnny (changes requested):

src/components/[guild]/Requirements/components/GuildCheckout/components/FeesTable.tsx
Comment (dominik-stumpf) [reply src/components/[guild]/Requirements/components/GuildCheckout/components/FeesTable.tsx:56]: Actually I'm not sure if we should change the `FeesTable` component or create a new one specifically for dynamic rewards. I can also try to make only the first column break word (on dynamic rewards only). What do you think @BrickheadJohnny ?

Comment (BrickheadJohnny) [reply src/components/[guild]/Requirements/components/GuildCheckout/components/FeesTable.tsx:56]: I think we can keep using the `FeesTable` component (maybe we could find a more general name for it, but we can leave it as is for now), just make the 1st column break word 👍

BrickheadJohnny (approved):

PR #1297, opened 2024-06-19
chore: add env variable validation

dominik-stumpf:
Comments starting with `WARNING: ` were placed on non-existent environment variables. How should we proceed with them? @BrickheadJohnny


src/pages/[guild]/collect/[chain]/[address].tsx
```diff
@@ -281,6 +279,7 @@ const getStaticProps = async ({ params }) => {
 
   // Calling the serverless endpoint, so if we fetch this data for the first time, it'll be added to the Vercel cache
 
+  // WARNING: this environment variable probably doesn't exist
```
Comment (BrickheadJohnny) [inline src/pages/[guild]/collect/[chain]/[address].tsx:282]: Actually it's a [Vercel system environment variable](https://vercel.com/docs/projects/environment-variables/system-environment-variables), so it won't exist on localhost, only on Vercel. I think it's save to use it this way here (without including it in our custom env var validator) 👌

src/wagmiConfig/index.ts
```diff
@@ -184,6 +185,7 @@ export const wagmiConfig = createConfig({
   connectors: process.env.NEXT_PUBLIC_MOCK_CONNECTOR
     ? [
         mock({
+          // WARNING: environment variable probably doesn't exist
```
BrickheadJohnny, line 188: Yeah, well, it won't exist on Vercel, we only use it locally & on GitHub for running our E2E tests.

BrickheadJohnny (commented):
Left 2 short comments & opened a PR with some cleanup/suggestions: https://github.com/guildxyz/guild.xyz/pull/1299 😉

BrickheadJohnny (approved):

PR #1298, opened 2024-06-19
Add 10 chains for NFT reward

TomiOhl:
### General details

- Linear issue: [GUILD-3245](https://linear.app/guildxyz/issue/GUILD-3245/deploy-nft-reward-to-many-chains)
- Discord thread: [forum > NFT Minting updates](https://discord.com/channels/697041998728659035/1225905536554500326)

### Core implementation details

Added the chains where necessary, no es mucho pero es trabajo honesto

### Notes

- needs #1293 to be merged first, that's why is the CI red
- there might be alchemy support for Arbitrum for example?


BrickheadJohnny (approved):
wow 🔥

src/pages/api/nft/collectors/[chain]/[address].ts
```diff
@@ -32,6 +32,16 @@ export const alchemyApiUrl: Record<ContractCallSupportedChain, string> = {
   ZKSYNC_ERA: "",
   LINEA: "",
   CYBER: "",
+  ARBITRUM: "",
```
Comment (TomiOhl) [inline src/pages/api/nft/collectors/[chain]/[address].ts:35]: Note that in guild-gate we are using `https://arb-mainnet.g.alchemy.com/nft/v2/{apikey}`, so we might be able to add an alchemy link here, but for that I will need help @BrickheadJohnny

Comment (BrickheadJohnny) [reply src/pages/api/nft/collectors/[chain]/[address].ts]: Oh I see, we usually separate projects (core - gate - frontend) for some reason on Alchemy, but actually the frontend won't call these endpoints frequently, so I think it's okay to add it here. I implemented the necessary change in https://github.com/guildxyz/guild.xyz/pull/1298/commits/ea0053a9f31d58985e3475e5cf54b854d409994c - we'll just need to add the necessary environment variable on Vercel. 👌

PR #1307, opened 2024-06-20
feat: add link to farcaster user requirement

dominik-stumpf:
Datablock replaced by link. I wonder if we should handle the undefined `data` cases when it isn't loaded.

### DIFF
--- a/src/requirements/Farcaster/FarcasterRequirement.tsx
+++ b/src/requirements/Farcaster/FarcasterRequirement.tsx
@@ -91,7 +91,7 @@ const FarcasterFollowChannel = (props: RequirementProps) => {
   return (
     <Requirement image={REQUIREMENTS.FARCASTER.icon.toString()} {...props}>
       {"Follow the "}
-      <Skeleton isLoaded={!!data}>
+      <Skeleton isLoaded={!!farcasterChannel}>
         <Link
           href={`https://warpcast.com/~/channel/${data?.id}`}
           isExternal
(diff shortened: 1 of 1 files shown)


BrickheadJohnny:
> Datablock replaced by link. I wonder if we should handle the undefined `data` cases when it isn't loaded.

Yup, please handle those too!

src/requirements/Farcaster/FarcasterRequirement.tsx
```diff
@@ -87,15 +91,17 @@ const FarcasterFollowChannel = (props: RequirementProps) => {
...
-      <Link
-        href={`https://warpcast.com/~/channel/${data.id}`}
-        isExternal
-        colorScheme="blue"
-        fontWeight="medium"
-      >
-        {farcasterChannel?.label ?? data.id}
-        <Icon as={ArrowSquareOut} mx="1" />
-      </Link>
+      <Skeleton isLoaded={!!data}>
```
BrickheadJohnny, line 94: I think we need to use `farcasterChannel` here

src/requirements/Farcaster/FarcasterRequirement.tsx
```diff
@@ -112,7 +118,9 @@ const FarcasterIncludeText = (props: RequirementProps) => {
   return (
     <Requirement image={REQUIREMENTS.FARCASTER.icon.toString()} {...props}>
       <Text as="span">{"Have "}</Text>
-      <DataBlockWithCopy text={data.id} />
+      <Skeleton isLoaded={!!data}>
```
BrickheadJohnny, line 121: `data` will always be loaded (it's the part of the requirement), so we don't need a skeleton here

BrickheadJohnny (changes requested):
Left 2 minor comments, feel free to merge once you fix those! 🚀

PR #1309, opened 2024-06-20
feat: simplify the create guild flow

BrickheadJohnny:
Basically removed the old multi-step solution & now we only have 3 inputs:
- Image (optional, if the user uploads one, we set their guild's `theme.color` to one which matches the image's color)
- Name (required)
- Email (required, bit pre-filled when possible)

I also removed the guild creation flow related E2E tests & the `GuildFormType` type, replaced it everywhere with types from the `@guildxyz/types` package.

We said that we could include a Twitter share button in the success toast, but we'll actually replace all those buttons with Farcaster cast buttons, so I haven't added it yet, will make sure to add it next week.


src/components/create-guild/CreateGuildForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+import CreateGuildButton from "./CreateGuildButton"
+import IconSelector from "./IconSelector"
+import Name from "./Name"
+
+export type CreateGuildFormType = Pick<
+  Schemas["GuildCreationPayload"],
+  "name" | "imageUrl" | "contacts" | "theme"
+>
+
+const getColorByImage = (imageUrl: string) =>
```
dovalid, line 20: I think it'd be nice to move it into a util function, and delete useSetImageAndNameFromPlatformData as it's not used anymore

dovalid (approved):

PR #1310, opened 2024-06-21
Copy improvements

BrickheadJohnny:
I'm not really sure about the wallet activity requirement copy, so let me know if you have any suggestions!

### DIFF
--- a/src/requirements/WalletActivity/WalletActivityRequirement.tsx
+++ b/src/requirements/WalletActivity/WalletActivityRequirement.tsx
@@ -36,7 +36,7 @@ const WalletActivityRequirement = (props: RequirementProps): JSX.Element => {
     if (maxAmount && minAmount === undefined)
       return (
         <>
-          {`Have a wallet created before `}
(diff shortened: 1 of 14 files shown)


src/requirements/WalletActivity/WalletActivityRequirement.tsx
```diff
@@ -36,30 +36,30 @@ const WalletActivityRequirement = (props: RequirementProps): JSX.Element => {
...
+          {`Have a transaction before `}
           <DataBlockWithDate timestamp={maxAmount} />
         </>
       )
 
     if (maxAmount === undefined && minAmount)
       return (
         <>
-          {`Have a wallet created after `}
+          {`Have a transaction after `}
```
TomiOhl, line 47: Since this is wallet age, maybe this would be more correct?
```suggestion
          {`Have your first transaction after `}
```
(And the other ones similarly)

dovalid (approved):
I agree with the suggestions of Tomi & Bruno, other than that, approved!

PR #1311, opened 2024-06-21
feat: add fallback image to reward icon

dominik-stumpf:


src/components/[guild]/RoleCard/components/Reward.tsx
```diff
@@ -191,6 +191,10 @@ export type RewardIconProps = {
 const MotionImg = motion(Img)
 const MotionCircle = motion(Circle)
 
+const RewardIconFallbacks = {
+  discord: "platforms/discord.png",
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RoleCard/components/Reward.tsx:195]: I think it would make sense to just use the icon from the reward config. So `rewards[PlatformType[guildPlatform.platformId]].imageUrl`. What do you think about this approach?

src/components/[guild]/RoleCard/components/Reward.tsx
```diff
@@ -197,53 +199,62 @@ const RewardIcon = ({
...
   }
+  const circleBgColor = useColorModeValue("gray.700", "gray.600")
+  const circleProps = {
+    bgColor: circleBgColor,
+    boxSize: 6,
+  }
 
-  if (!props.src) {
-    if (withMotionImg)
+  if (doIconFallback) {
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RoleCard/components/Reward.tsx:225]: I think we could keep the `!props.src` condition here too, what do you think? (In that case users wouldn't need to fetch the image, then fallback to the icon if it errors out)

BrickheadJohnny (approved):

PR #1314, opened 2024-06-24
shadcn/ui migration `/explorer`

dominik-stumpf:


BrickheadJohnny:
Let's break this down into smaller chunks in https://github.com/guildxyz/guild.xyz/pull/1323

PR #1318, opened 2024-06-25
refactor: remove the custom `Requirement` type

BrickheadJohnny:


src/components/[guild]/Requirements/components/RequirementDisplayComponent.tsx
```diff
@@ -22,7 +22,7 @@ const RequirementDisplayComponent = ({
   rightElement = <RequirementAccessIndicator />,
   ...rest
 }: Props) => {
-  if (requirement.type === "HIDDEN")
+  if (requirement.visibility === "HIDDEN")
```
Comment (BrickheadJohnny) [inline src/components/[guild]/Requirements/components/RequirementDisplayComponent.tsx:25]: I wasn't sure about this change, let me know if you have a better idea! It's needed so we can display "Some secret requirements" in the requirements list.

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/SelectExistingPlatform.tsx
```diff
@@ -36,7 +30,7 @@ const SelectExistingPlatform = ({ onClose, onSelect }: Props) => {
     name: "rolePlatforms",
   })
 
-  const roleVisibility: Visibility = useWatch({ name: ".visibility" })
+  const roleVisibility = useWatch({ name: ".visibility" })
```
Comment (cs-balazs) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/SelectExistingPlatform.tsx:33]: Should we remove the leading `.` here as well?

src/requirements/Allowlist/AllowlistRequirement.tsx
```diff
@@ -23,7 +23,8 @@ function HiddenAllowlistText({ isEmail }: { isEmail: boolean }) {
 }
 
 const AllowlistRequirement = ({ ...rest }: RequirementProps): JSX.Element => {
-  const requirement = useRequirementContext() as Extract<
+  const requirement = useRequirementContext<"ALLOWLIST" | "ALLOWLIST_EMAIL">()
+  const castedRequirement = requirement as unknown as Extract<
```
cs-balazs, line 27: This might be worth a TODO comment, that we should add `addressCount` to the schema, and remove the cast later on

src/components/[guild]/Requirements/RequirementErrorConfigContext.tsx
```diff
@@ -5,7 +5,7 @@ type RequirementErrorConfig = Partial<Record<RequirementType, string>>
 
 const REQ_ERROR_CONFIG: RequirementErrorConfig = {
   GALAXY: "Galxe API error, please try again later",
-  TWITTER: "X API error, please try again later",
+  // TWITTER: "X API error, please try again later",
```
Comment (cs-balazs) [inline src/components/[guild]/Requirements/RequirementErrorConfigContext.tsx:8]: Why has this been commented? Typing-related, or we don't need it?

cs-balazs (approved):
Fantastic changes! Glad to see these schemas are being used more 🚀

src/components/[guild]/Requirements/components/RequirementDisplayComponent.tsx
Comment (cs-balazs) [reply src/components/[guild]/Requirements/components/RequirementDisplayComponent.tsx:25]: I'm honestly not sure why we checked `type` here. I saw, we even had `| "HIDDEN"` in our previous type. I'd say, if the functionality still works, it should be fine :ayy:

Comment (BrickheadJohnny) [reply src/components/[guild]/Requirements/components/RequirementDisplayComponent.tsx:25]: Yeah, we actually assigned `HIDDEN` to the type somewhere else, I guess that was a "temporary" workaround :'D

src/components/[guild]/Requirements/RequirementErrorConfigContext.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/Requirements/RequirementErrorConfigContext.tsx:8]: Typing-related & we don't really need it anymore. :D Will remove it in another PR!

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/SelectExistingPlatform.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/SelectExistingPlatform.tsx]: Double checked and it works correctly without the dot, so removed it, thx for pointing it out!

PR #1322, opened 2024-06-26
[GUILD-3374] - Farcaster: follow, like and recast actions

cs-balazs:
[Linear](https://linear.app/guildxyz/issue/GUILD-3374/follow-like-recast-requirement-button)

- Adds 3 possible requirement buttons
  - `FARCASTER_FOLLOW`: If the user has a connected Farcaster account, and doesn't have access, we show a follow button, which performs the Farcaster follow, then rechecks the role
  - Similarly like/recast for `FARCASTER_LIKE` and `FARCASTER_RECAST`
  - The submit logic is handled by `useFarcasterAction`. The two button components look somewhat similar, and we probably could refactor it to only have one, but in the future, we will have more actions, that might not look this similar, so I left it like this. Let me know what you think about this!
- Made some changes to make TSC happy: https://github.com/guildxyz/guild.xyz/pull/1322/commits/3b3a143b159440de1af275a6680d007b250586e7
- For some reason the `Skeleton`-s looked wrong, they needed a `display="inline"`. This might be a more general issue, I just fixed it where I encountered it while testing


src/requirements/Farcaster/components/FarcasterActionFollow.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  const { data, roleId } = useRequirementContext<
+    "FARCASTER_FOLLOW" | "FARCASTER_FOLLOWED_BY"
+  >()
+
+  const { data: farcasterUser } = useFarcasterUser(data.id)
+
+  const { onSubmit, isLoading, response } = useFarcasterAction(roleId, "follow")
+
+  if (!farcasterProfiles || !farcasterProfiles?.[0]?.fid || response) {
```
BrickheadJohnny, line 18: Do we actually need this first condition for some reason? Isn't the optional chaining enough for this check? 👀

src/requirements/Farcaster/hooks/useFarcasterAction.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const toast = useToast()
+  const showErrorToast = useShowErrorToast()
+
+  return useSubmit(
+    async (param?: string) => {
+      if (!param) {
+        throw new Error("Unknown Farcaster action")
+      }
+      if (!farcasterProfiles?.[0]?.fid) {
+        throw new Error("First connect ypur Farcaster account")
```
BrickheadJohnny, line 25: tpyo :kappa:

BrickheadJohnny (approved):
Awesome PR! Left just 2 minor comments.
I agree that we could merge the action buttons (and maybe they wouldn't even need to accept params, we could just get the necessary data from the requirement context & use a config object to render the appropriate labels, etc. on them?), but they don't look messy this way either, so it's not that important. 😉

src/requirements/Farcaster/hooks/useFarcasterAction.tsx
cs-balazs: Nice catch! Fixed in https://github.com/guildxyz/guild.xyz/pull/1322/commits/58cd9a648cadb7849db6b792616294d5e296d439

src/requirements/Farcaster/components/FarcasterActionFollow.tsx
cs-balazs: That's right, removed these checks in https://github.com/guildxyz/guild.xyz/pull/1322/commits/3f6cac54c1ca0ab3850fe48ec42bea2a0625a326

PR #1331, opened 2024-06-27
feat(activity log): don't fetch `click join on web` actions

BrickheadJohnny:
The `join guild` action was a child action of `click join on web` before, but now it's also a parent action, so we can use the former one. The original `click join on web` was a bit misleading, since it is triggered for access checks too.

Relevant backend PR: https://github.com/guildxyz/guild-backend/pull/1529


cs-balazs (approved):
I'm not very familiar with this code, but the changes make sense to me 🚀

PR #1332, opened 2024-06-28
Storybook

BrickheadJohnny:
- Configured Storybook & added stories for our `Button` component
- Added new CSS variables for hover & active colors, because in my opinion, changing only the opacity of a color in these states isn't the best idea. It certainly won't work for custom colors (e.g. for the generated colors on the guild pages)
- Removed the button examples from the playground page, we'll eventually use Storybook only & remove that page before merging the main PR
- Removed the local Inter font


.storybook/preview.ts
```diff
@@ -0,0 +1,31 @@
+import type { Preview } from "@storybook/react"
+import "../src/app/globals.css"
+import { dystopian, inter } from "../src/fonts"
+
+import { withThemeByDataAttribute } from "@storybook/addon-themes"
+
+document?.querySelector("body")?.classList.add(inter.variable, dystopian.variable)
```
BrickheadJohnny, line 7: Not sure if this is the best solution...

BrickheadJohnny:
Merging it, because I need it for #1329, but I'd like to get some feedback on it later!

PR #1333, opened 2024-07-01
feat: `Toast` component & `useToast` hook

BrickheadJohnny:
Implemented a simple toast component. I know we sometimes add icons to our toasts, but for now I haven't implemented that functionality, we can add it once we actually need it.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -104,11 +104,6 @@
     --toast-error: 317 4% 35%;
     --toast-warning: 30 6% 34%;
     --toast-info: 226 7% 36%;
-
-    --sm: 640px;
-    --md: 768px;
-    --lg: 1024px;
-    --xl: 1280px;
   }
 
   :root[data-theme="dark"] video,
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -46,10 +46,10 @@
     --input: var(--border);
     --ring: 224 71.4% 4.1%;
 
-    --toast-success: 141, 79%, 85%, 0.16;
-    --toast-error: 0, 96%, 89%, 0.16;
-    --toast-warning: 38, 93%, 77%, 0.16;
-    --toast-info: 213, 97%, 87%, 0.16;
+    --toast-success: 137 64% 98%;
+    --toast-error: 0 78% 98%;
+    --toast-warning: 40 79% 96%;
+    --toast-info: 218 80% 98%;
 
     --sm: 640px;
     --md: 768px;
@@ -100,10 +100,10 @@
     --input: var(--border);
     --ring: 240 0% 45%;
 
-    --toast-success: 141, 79%, 85%, 0.16;
-    --toast-error: 0, 96%, 89%, 0.16;
-    --toast-warning: 38, 93%, 77%, 0.16;
-    --toast-info: 213, 97%, 87%, 0.16;
+    --toast-success: 180 4% 35%;
+    --toast-error: 317 4% 35%;
+    --toast-warning: 30 6% 34%;
+    --toast-info: 226 7% 36%;
 
     --sm: 640px;
     --md: 768px;
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -45,6 +45,16 @@
     --border: 240 6% 90%;
     --input: var(--border);
     --ring: 224 71.4% 4.1%;
+
+    --toast-success: 141, 79%, 85%, 0.16;
+    --toast-error: 0, 96%, 89%, 0.16;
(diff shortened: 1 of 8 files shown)


src/app/globals.css
```diff
@@ -45,6 +45,16 @@
...
+
+    --toast-success: 137 64% 98%;
+    --toast-error: 0 78% 98%;
+    --toast-warning: 40 79% 96%;
+    --toast-info: 218 80% 98%;
+
+    --sm: 640px;
+    --md: 768px;
+    --lg: 1024px;
+    --xl: 1280px;
```
dominik-stumpf, line 57: The dark theme shouldn't include (override) these values

src/v2/components/ui/Toast.tsx
```diff
@@ -0,0 +1,122 @@
...
+      "fixed right-0 top-0 z-[100] flex max-h-screen w-full max-w-[min(420px,_100%)] flex-col gap-2 p-4 focus:outline-none",
+      className
+    )}
+    {...props}
+  />
+))
+ToastViewport.displayName = ToastPrimitives.Viewport.displayName
+
+const toastVariants = cva(
+  "bg-card text-foreground group pointer-events-auto relative flex w-full items-center justify-between space-x-4 overflow-hidden rounded-md p-4 pr-8 shadow-lg transition-transform data-[swipe=cancel]:translate-x-0 data-[swipe=end]:translate-x-[var(--radix-toast-swipe-end-x)] data-[swipe=move]:translate-x-[var(--radix-toast-swipe-move-x)] data-[swipe=move]:transition-none data-[state=open]:animate-in data-[state=closed]:animate-out data-[swipe=end]:animate-out data-[state=closed]:fade-out-80 data-[state=closed]:slide-out-to-right-full data-[state=open]:slide-in-from-right-full data-[state=open]:sm:slide-in-from-right-full focus:outline-none",
```
dominik-stumpf, line 28: We should add a `focus-visible:ring-2 focus-visible:ring-ring` to highlight the toast itself on keyboard navigation (first place where tab brings the user to). What do you think?

src/v2/components/ui/hooks/useToast.ts
```diff
@@ -0,0 +1,189 @@
+"use client"
+
+import type { ToastActionElement, ToastProps } from "@/components/ui/Toast"
+import { ReactNode, useEffect, useState } from "react"
+
+const TOAST_LIMIT = 3
+const TOAST_REMOVE_DELAY = 1000000
```
dominik-stumpf, line 7: Later maybe we should make this accessible from the outside? I'm thinking it would be useful if the storybook version would sustain and could be inspectable.

src/v2/components/ui/hooks/useToast.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    case "DISMISS_TOAST": {
+      const { toastId } = action
+
+      // ! Side effects ! - This could be extracted into a dismissToast() action,
+      // but I'll keep it here for simplicity
+      if (toastId) {
+        addToRemoveQueue(toastId)
+      } else {
+        state.toasts.forEach((toast) => {
```
dominik-stumpf, line 96: Interestingly there is a shadow for `toast` param here conflicting with line 140, should we follow eslint errors? There are other instances.

src/v2/components/ui/hooks/useToast.ts
BrickheadJohnny: Yup, that's a great idea! Didn't want to spend too much time with it, since we may replace it with Sonner soon! ;)

src/app/globals.css
BrickheadJohnny: Ah, I wasn't sure if it overrides it, and then forgot to test it, thanks, removed in https://github.com/guildxyz/guild.xyz/pull/1333/commits/f9f02fad29940fb4d782cc0068eadf7fa13cce10

src/v2/components/ui/Toast.tsx
BrickheadJohnny: Added in https://github.com/guildxyz/guild.xyz/pull/1333/commits/6b7a3cfa32b49fe35e290de11051f89e9fb1dd77

src/v2/components/ui/hooks/useToast.ts
BrickheadJohnny: Nice catch, fixed in https://github.com/guildxyz/guild.xyz/pull/1333/commits/a7a0a993fcd55c42cb6a2424e4a462f1873ca6bb

PR #1335, opened 2024-07-01
feat: add tooltip

dominik-stumpf:


BrickheadJohnny (approved):
I quickly added the package to `package.json`, updated the styles, and added a simple story for it. ;) I removed the arrow, it's not that important, but we can add it later. Let's merge this!

PR #1340, opened 2024-07-02
refactor(useCreateRRR): feedback configuration

FBalint:


dovalid (commented):
Are you sure we need this much configuration? I don't see a case where we'd want to display multiple success toasts / confetti, or when we wouldn't want to show an error if it occoured. So I think the right way is to just have the good old onSuccess prop for the final onSubmit as we do have now, remove the abstracted toast & confetti from that, and pass them always as we want. I'm happy to talk about it tomorrow on voice!

FBalint:
> Are you sure we need this much configuration? I don't see a case where we'd want to display multiple success toasts / confetti, or when we wouldn't want to show an error if it occoured. So I think the right way is to just have the good old onSuccess prop for the final onSubmit as we do have now, remove the abstracted toast & confetti from that, and pass them always as we want. I'm happy to talk about it tomorrow on voice!

Maybe I got a bit ahead of myself with overengineering, so I simplified it a great deal

dovalid (approved):

PR #1348, opened 2024-07-03
resolve circular deps inside rewards

dominik-stumpf:
# Fixes 2 circular dependency group

- **fix(SecretText): resolve circular dependency inside rewards**
- **fix(PolygonID): resolve circular dependency inside rewards**

### DIFF
--- a/src/rewards/PolygonID/components/ConnectDIDModal.tsx
+++ b/src/rewards/PolygonID/components/ConnectDIDModal.tsx
@@ -23,18 +23,19 @@ import { useEffect } from "react"
 import useSWR from "swr"
 import useSWRImmutable from "swr/immutable"
 import useConnectedDID from "../hooks/useConnectedDID"
-import { useMintPolygonIDProofContext } from "./MintPolygonIDProofProvider"
 
 type Props = {
   isOpen: boolean
   onClose: () => void
+  onMintPolygonIDProofModalOpen: () => void
 }
 
-const ConnectDIDModal = ({ isOpen, onClose }: Props) => {
+const ConnectDIDModal = ({
+  isOpen,
+  onClose,
+  onMintPolygonIDProofModalOpen,
+}: Props) => {
   const { id: userId } = useUser()
-
-  const { onMintPolygonIDProofModalOpen } = useMintPolygonIDProofContext()
-
   const toast = useToast()
 
   const fetcherWithSign = useFetcherWithSign()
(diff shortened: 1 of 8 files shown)


src/rewards/PolygonID/components/ConnectDIDModal.tsx
```diff
@@ -23,18 +23,19 @@ import { useEffect } from "react"
...
   isOpen: boolean
   onClose: () => void
+  onMintPolygonIDProofModalOpen: () => void
 }
 
-const ConnectDIDModal = ({ isOpen, onClose }: Props) => {
+const ConnectDIDModal = ({
+  isOpen,
+  onClose,
+  onMintPolygonIDProofModalOpen,
```
BrickheadJohnny, line 36: What was the root of the issue here? We used a context so we don't need to pass down props like this, so it would be neat if we could solve this in some other way :D

src/rewards/SecretText/SecretTextDataForm/types.ts
```diff
@@ -0,0 +1,5 @@
+export type SecretTextRewardForm = {
```
BrickheadJohnny, line 1: I'm not sure if we need to put these types in a `types.ts` close to the components where we actually use them, or we should just put them in the main `src/types.ts` file. What are the pros and cons in your opinion?

BrickheadJohnny (commented):
I had 2 questions, let's discuss them 😉

src/rewards/SecretText/SecretTextDataForm/types.ts
dominik-stumpf: Pros:
- The `src/types.ts` won't grow unmaintainably (name collision, too large file, uncertain types)
- The types will be localized in some sense
- This way it's clear that the `types.ts` belongs to `SecretText` from every import, this gives the type additional data for the developers.

Cons:
- More files
- Global types doesn't require imports, as this one does so when something is renamed or moved, more refactor is required usually

I plan to write about these in the project structure convention, this is not something we do in the project now.

src/rewards/PolygonID/components/ConnectDIDModal.tsx
dominik-stumpf: This component isn't used anywhere, but in that provider. Other solution perhaps could be if we moved this into the provider.

BrickheadJohnny: > This component isn't used anywhere, but in that provider.
Oh, I missed that during the review, then it's great this way!

BrickheadJohnny (approved):

PR #1349, opened 2024-07-03
fix(Farcaster): resolve circular dependency

dominik-stumpf:
# Fixes 1 circular dependency group

### DIFF
--- a/src/requirements/Farcaster/types.ts
+++ b/src/requirements/Farcaster/types.ts
@@ -1,0 +1,11 @@
+export type FarcasterCastData = {
+  hash: string
+  username: string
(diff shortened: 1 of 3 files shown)


src/requirements/Farcaster/types.ts
```diff
@@ -0,0 +1,11 @@
+export type FarcasterCastData = {
```
BrickheadJohnny, line 1: Left a comment under your other PR about this, but LGTM actually!

BrickheadJohnny (approved):

BrickheadJohnny (approved):

PR #1351, opened 2024-07-03
`Dialog` improvements

BrickheadJohnny:
Made the overlay scrollable & added a Dialog story so we can easily preview all sizes / variants.


dominik-stumpf (approved):
:rocket:

PR #1353, opened 2024-07-04
[GUILD-3399] - Farcaster connection can reuse existing approved keys

cs-balazs:
- It is now possible, that a Farcaster connection doesn't need a QR code, instead connects an existing key to the user
- This PR handles this case, when we don't need to show a QR code
- BE: https://github.com/guildxyz/guild-backend/pull/1541


BrickheadJohnny (approved):
🚀

PR #1356, opened 2024-07-04
Update Viem, Wagmi, add Mint chain

marclllaks:


BrickheadJohnny (approved):
very good very nice 🚀

PR #1357, opened 2024-07-04
resolve circular deps

dominik-stumpf:


BrickheadJohnny (approved):

src/components/[guild]/RoleCard/components/Reward.tsx
Comment (BrickheadJohnny) [inline src/components/[guild]/RoleCard/components/Reward.tsx:1]: At this point we could maybe move the `RewardIcon` component to its own file too, wdyt?

BrickheadJohnny (approved):
LGTM, but left 1 question/suggestion!

src/components/[guild]/RoleCard/components/Reward.tsx
Comment (dominik-stumpf) [reply src/components/[guild]/RoleCard/components/Reward.tsx:1]: I tried that too, but since it uses reward it wouldn't make a difference (for the case of circular imports).

Comment (BrickheadJohnny) [reply src/components/[guild]/RoleCard/components/Reward.tsx:1]: Ah I see, but we use it in other places where we don't use the `Reward` component itself, so maybe we could avoid importing unnecessary stuff in those places? 👀

Comment (dominik-stumpf) [reply src/components/[guild]/RoleCard/components/Reward.tsx:1]: Well, I think if `Reward.tsx` wouldn't import `reward` or `rewardComponent` the component itself could be cleaner where the indexing is handled in higher components. If so, it wouldn't be a problem anymore to extract `RewardIcon`

PR #1363, opened 2024-07-09
feat: add compound layout component

dominik-stumpf:
Blocked by #1329


src/v2/components/Layout/Layout.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const Root = ({ children }: PropsWithChildren) => (
+  <div className="flex min-h-screen flex-col">{children}</div>
+)
+
+/* -------------------------------------------------------------------------------------------------
+ * Header
+ * -----------------------------------------------------------------------------------------------*/
+
+const Header = ({ children }: PropsWithChildren) => (
```
BrickheadJohnny, line 19: Let's call this `Hero` & don't include a `<header>` tag here, since we already use one inside our `Header` component

src/v2/components/Layout/Layout.tsx
```diff
@@ -0,0 +1,76 @@
...
+      {title}
+    </h1>
+  </PageBoundary>
+)
+
+/* -------------------------------------------------------------------------------------------------
+ * Banner
+ * -----------------------------------------------------------------------------------------------*/
+
+const Banner = () => (
```
BrickheadJohnny, line 46: This component should only handle the size & position (and it could have a default background color) & we should pass the customization related stuff to it where we actually use it ;)

src/v2/components/Layout/Layout.tsx
```diff
@@ -0,0 +1,76 @@
...
+  </div>
+)
+
+/* -------------------------------------------------------------------------------------------------
+ * Main
+ * -----------------------------------------------------------------------------------------------*/
+
+const Main = ({ children }: PropsWithChildren) => (
+  <main>
+    <PageBoundary>{children}</PageBoundary>
```
BrickheadJohnny, line 60: `PageBoundary` could support the `asChild` prop so we wouldn't need to always nest components like this.

src/v2/components/PageBoundary.tsx
```diff
@@ -0,0 +1,12 @@
+import clsx from "clsx"
+import { forwardRef } from "react"
+
+export type PageBoundaryProps = React.HTMLAttributes<HTMLDivElement>
+export const PageBoundary = forwardRef<HTMLDivElement, PageBoundaryProps>(({ children, className }, ref) => (
```
BrickheadJohnny, line 5: nit: `Container` sounds better in my opinion

BrickheadJohnny (changes requested):

PR #1365, opened 2024-07-10
ci: add nexjs-bundle-analysis

dominik-stumpf:


dominik-stumpf:
Building without env vars proves to be problematic and this feature isn't that useful anyway, closing.

PR #1375, opened 2024-07-11
ci: add tsc to pre-commit hook

dominik-stumpf:


.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx lint-staged
+npm run type-check -- --incremental false
```
dovalid, line 5: Why is the modifier needed, don't we want to run it incrementally so it's faster?

dovalid (approved):

PR #1383, opened 2024-07-13
chore: import all icons from direct paths

BrickheadJohnny:
Checking if we can fix the "Too many open files" issue on Vercel.


BrickheadJohnny:
At least we tried. 🤷

PR #1388, opened 2024-07-15
font fix with direct and optimize import

dominik-stumpf:
Attempt icon fix, branching from #1383

- **chore: import all icons from direct paths**
- **fix(_error): import missing icon**
- **fix: import issues**


dominik-stumpf:
This issue was fixed by #1391. Closing!

PR #1396, opened 2024-07-16
add metadata to page

dominik-stumpf:


src/app/layout.tsx
```diff
@@ -14,8 +13,20 @@ interface RootLayoutProps {
...
   description:
     "Automated membership management for the platforms your community already uses.",
+  icons: {
+    icon: "guild-icon.png",
+  },
+}
+
+export const viewport: Viewport = {
+  themeColor: [
+    { media: "(prefers-color-scheme: dark)", color: "#000000" },
```
BrickheadJohnny, line 26: I think we could use the current dark gray background color here and in the `manifest.ts` file too (`#27272a`)

BrickheadJohnny (approved):

PR #1399, opened 2024-07-17
resolve circular dependencies

dominik-stumpf:
This PR resolves 6 circular import errors. Loops were addressed separately by each *dcd* commit.


src/requirements/Form/components/ExpectedAnswerCardView.tsx
BrickheadJohnny, line 1: We could call it `ExpectedAnswerCardDisplay` for consistency. Or, we could rename all components inside it to `Expected...UI` and call it `ExpectedAnswerCardUI`.

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddTokenPanel/components/TokenAmountStep.tsx
```diff
@@ -34,10 +34,10 @@ const TokenAmountStep = ({ onContinue }: { onContinue: () => void }) => {
     (gp) => gp.platformId === PlatformType.ERC20
   )
 
-  const platformForToken = tokenPlatforms.find(
+  const platformForToken = tokenPlatforms?.find(
     (guildPlatform) =>
-      guildPlatform.platformGuildData.chain === chain &&
-      guildPlatform.platformGuildData.tokenAddress.toLowerCase() ===
+      guildPlatform.platformGuildData?.chain === chain &&
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddTokenPanel/components/TokenAmountStep.tsx:39]: We should somehow know that `platformGuildData` is defined at this point, so not sure about this change. I know that TS warned us here, but in reality, this field will be defined here hehe.

BrickheadJohnny (approved):
Left 2 minor comments, but otherwise LGTM

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddTokenPanel/components/TokenAmountStep.tsx
Comment (dominik-stumpf) [reply src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddTokenPanel/components/TokenAmountStep.tsx]: Agreed. Reverted the optional chaining, but at this point the `useGuild` hook caused so many type errors like this. Maybe we should attempt some type wizardry there? @BrickheadJohnny

Comment (BrickheadJohnny) [reply src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddTokenPanel/components/TokenAmountStep.tsx]: Yes, actually we should just start using the `GuildReward` type [from our types package](https://github.com/guildxyz/guild-backend/blob/main/lib/types/schemas/guildReward.ts), that would allow us to narrow down the type & TS would automatically know at this point that `platformGuildData` is defined & it'd also know its properties... Let's talk about it next week! 😉

PR #1402, opened 2024-07-18
fix: add existing rewards to a guild

BrickheadJohnny:
Users sometimes do not notice the rewards which they already use in their guild in the quick add section, so we display these rewards in components like `DiscordGuildSetup` too. We forgot to handle & test this case in a recent refactor, so this feature caused some errors. This PR aims to fix the issue by checking if a `guildPlatform` with a specific `platformGuildId` already exists in the guild, and optionally adding a `guildPlatformId` field to the `rolePlatform` in that case, so our API knows that we shouldn't try to create a new `guildPlatform` in our DB, but we should re-use an existing one. I added a longer comment about it too. We also display a settings modal when the user wants to add an existing reward to a role. I haven't added this modal to the `AddSolutionsModal`, because in my opinion users will probably click on that button if they want to add a brand new reward to their guild, but let me know if we should add it there too.

As a side quest, I completely removed some unused files & added the new solutions modal to the group page too.


dovalid (approved):

dovalid (approved):
I agree with you about adding the settings modal to the AddSolutionModal, I'm sure we're good this way for now

PR #1403, opened 2024-07-18
Implement `/profile` presentation and container

dominik-stumpf:
~~PR responsible for implementing the `/profile` route design and preparing abstracted components later to be joined up with the backend by another PR.~~ This PR is responsible for collecting all features required by `/profile`


src/app/(marketing)/profile/[username]/page.tsx
```diff
@@ -0,0 +1,195 @@
...
+  params: { username: string }
+}) => {
+  const profile = await getProfileData(username)
+  const level = 50
+
+  return (
+    <Layout>
+      <LayoutHero>
+        <LayoutHeader />
+        <LayoutBanner className="-bottom-[500px]">
```
Comment (BrickheadJohnny) [inline src/app/(marketing)/profile/[username]/page.tsx:69]: Let's use the `Layout` component's capabilities here! The background image should be dynamic (its height basically depends on the profile description), so in my opinion we should remove this arbitrary value here and just put the image, name, handle, description, followers section inside `LayoutHero`, that way the CSS will handle the proper height. ;)

src/app/(marketing)/profile/[username]/page.tsx
```diff
@@ -0,0 +1,195 @@
...
+          <div className="absolute inset-0 bg-[url('/banner.svg')] opacity-10" />
+          <div className="absolute inset-0 bg-gradient-to-t from-background" />
+        </LayoutBanner>
+      </LayoutHero>
+      <LayoutMain>
+        <div className="mt-24">
+          <div className="relative mb-24 flex flex-col items-center">
+            <EditProfile
+              profileImageUrl={profile.profileImageUrl}
```
Comment (BrickheadJohnny) [inline src/app/(marketing)/profile/[username]/page.tsx:78]: What if we'd wrap the whole page in `SWRConfig` & provide the server side fetched data to it as a default value & create a `useProfile` hook to retrieve the necessary data inside each component? That way we wouldn't need to pass so many parameters here and in other places.

src/app/(marketing)/profile/[username]/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+  { color: "#88d525", title: "learner" },
+  { color: "#f6ca45", title: "knight" },
+  { color: "#78c93d", title: "veteran" },
+  { color: "#ec5a53", title: "champion" },
+  { color: "#53adf0", title: "hero" },
+  { color: "#c385f8", title: "master" },
+  { color: "#3e6fc3", title: "grand master" },
+  { color: "#be4681", title: "legend" },
+  { color: "#000000", title: "mythic" },
+  { color: "linear-gradient(to left top, #00cbfa, #b9f2ff)", title: "???" },
```
Comment (BrickheadJohnny) [inline src/app/(marketing)/profile/[username]/constants.ts:13]: ??? :D

src/app/(marketing)/profile/_components/EditContributions.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Schemas } from "@guildxyz/types"
+// import { zodResolver } from "@hookform/resolvers/zod"
+import { Pencil } from "@phosphor-icons/react"
+import { FormProvider, useForm } from "react-hook-form"
+
+export const EditContributions = (
+  contribution: Schemas["ProfileContributionUpdate"]
+) => {
+  const form = useForm<Schemas["ProfileContributionUpdate"]>({
+    // resolver: zodResolver(profileSchema),
```
BrickheadJohnny, line 22: Are there any issues with this? 👀

src/app/(marketing)/profile/_components/EditContributions.tsx
```diff
@@ -0,0 +1,61 @@
...
+  })
+
+  async function onSubmit(values: Schemas["ProfileContributionUpdate"]) {
+    console.log("edit contributions submit", values)
+  }
+
+  return (
+    <Dialog>
+      <DialogTrigger asChild>
+        <Button variant="solid" size="icon">
```
BrickheadJohnny, line 36: The button should be smaller & fully rounded (we can handle it here with a className & if we need to use the same button in multiple places, we could even add a variant to it)

src/app/(marketing)/profile/_components/EditContributions.tsx
```diff
@@ -0,0 +1,61 @@
...
+        </Button>
+      </DialogTrigger>
+      <DialogContent size="lg" className="bg-background">
+        <DialogHeader>
+          <DialogTitle>Edit profile</DialogTitle>
+          <DialogCloseButton />
+        </DialogHeader>
+        <FormProvider {...form}>
+          <form onSubmit={form.handleSubmit(onSubmit)}>
+            <DialogBody>
```
BrickheadJohnny, line 47: Will you implement the modal in this PR or you'll add it later in another one?

src/v2/components/ui/Textarea.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export interface TextareaProps
+  extends React.TextareaHTMLAttributes<HTMLTextAreaElement> {}
+
+const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
+  ({ className, ...props }, ref) => {
+    return (
+      <textarea
+        className={cn(
+          "flex min-h-[80px] w-full rounded-md border border-input bg-card-secondary px-3 py-2 text-sm ring-offset-background placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50",
```
BrickheadJohnny, line 13: I think this should have a default `rounded-lg` radius (like the default variant of the `Input` component). And the text shouldn't be small inside a textarea. (Just checked & we use small text inside inputs too, will make sure to change that too!)

Let's also modify the ring, it should look the same as on simple inputs.

src/app/(marketing)/profile/_components/OperatedGuildCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Card } from "@/components/ui/Card"
+import { Skeleton } from "@/components/ui/Skeleton"
+import { User } from "@phosphor-icons/react/dist/ssr"
+import { AvatarImage } from "@radix-ui/react-avatar"
+import Image from "next/image"
+
+export const OperatedGuildCard = () => {
+  return (
+    <Card className="flex">
+      <div className="relative w-1/3">
```
BrickheadJohnny, line 12: I think it would be easier to use grid here with 3 cols. That would also help to easily handle the mobile view which is missing right now 😉

src/app/(marketing)/profile/_components/OperatedGuildCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <div className="absolute inset-0 bg-black" />
```
BrickheadJohnny, line 13: We should add a horizontal padding to this container!

src/app/(marketing)/profile/_components/OperatedGuildCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+              <Skeleton className="size-full" />
+            </AvatarFallback>
+          </Avatar>
+          <h3 className="text-center font-bold font-display text-xl">
+            The Guilded Age Guild
+            <CheckMark className="ml-2 inline-block size-6" />
+          </h3>
+        </div>
+      </div>
+      <div className="grid grow grid-cols-2 justify-stretch gap-2 px-6 py-5">
```
BrickheadJohnny, line 39: Let's use 1 column in mobile view!

src/app/(marketing)/profile/_components/LevelBadge.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <div
+      className={cn("relative grid size-12 place-items-center shadow-lg", className)}
+    >
+      <div className="size-full rounded-full" style={{ background: rank.color }} />
+      <div className="absolute select-none font-extrabold text-xl">{level}</div>
+    </div>
+  )
+}
+// <Polygon sides={5} radius={20} color={"hsl(var(--primary))"} />
```
BrickheadJohnny, line 49: 🗑️

src/app/(marketing)/profile/_components/LevelBadge.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+type PolygonProps = {
+  sides: number
+  radius: number
+  color: string
+}
+
+const Polygon = ({ sides, radius, color }: PolygonProps) => {
+  const angleStep = (2 * Math.PI) / sides
+  const points = Array.from({ length: sides }, (_, i) => {
```
BrickheadJohnny, line 12: Wow, clever solution! 🤯

src/app/(marketing)/profile/_components/EditProfile.tsx
BrickheadJohnny, line 1: Inputs should be darker inside this modal (like in Figma)

src/app/(marketing)/profile/_components/EditProfile.tsx
```diff
@@ -0,0 +1,216 @@
...
+                          />
+                        ) : (
+                          <div className="size-full bg-purple-800" />
+                        )}
+                      </div>
+                      <div className="relative flex items-center gap-3">
+                        <Button size="icon" variant="ghost">
+                          <ImageIcon weight="bold" size={24} />
+                        </Button>
+                        <Separator orientation="vertical" className="h-6 w-0.5" />
```
BrickheadJohnny, line 120: The separator color is inaccurate, I think we can use an arbitrary value here (maybe a white color with 50% opacity), since this is an "edge case".

src/app/(marketing)/profile/_components/EditProfile.tsx
```diff
@@ -0,0 +1,216 @@
...
+                    <FormControl>
+                      <Textarea placeholder="" className="max-h-12" {...field} />
+                    </FormControl>
+                    <FormErrorMessage />
+                  </FormItem>
+                )}
+              />
+            </DialogBody>
+            <DialogFooter>
+              <Button colorScheme="destructive" type="submit">
```
BrickheadJohnny, line 200: This button is too strong in my opinion, I think we should try to use another variant. Maybe outline? And I'm not 100% sure if we actually need this functionality, because in theory if the user cancels their subscription, we just delete their profile, right? (Except for the lifetime subscription, I think)

src/app/(marketing)/profile/_components/RecentActivity.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { useState } from "react"
+
+export const RecentActivity = () => {
+  const ACTIVITY_FILTERS = ["All", "Editing", "Join", "NFTs"] as const
+  const [activityFilter, setActivityFilter] =
+    useState<(typeof ACTIVITY_FILTERS)[number]>("All")
+
+  return (
+    <>
+      <ToggleGroup
```
BrickheadJohnny, line 17: These toggles should be smaller according to Figma

BrickheadJohnny (commented):
This probably wasn't the final review, but I left a couple of comments for you! ;) Also, I believe that all "primary" colors (e.g. the charts, the progress around the profile and the rank badge) should be generated on the fly. We could actually generate several swatches from the current rank's color & use them.

src/app/(marketing)/profile/_components/LevelBadge.tsx
dominik-stumpf: At your service! :smile:

src/app/(marketing)/profile/[username]/constants.ts
Comment (dominik-stumpf) [reply src/app/(marketing)/profile/[username]/constants.ts:13]: Don't judge me. I'm just doing what I have to do. :laughing:

src/app/(marketing)/profile/_components/EditProfile.tsx
dominik-stumpf: This button is actually temporary, and only needed for development.

src/app/(marketing)/profile/_components/EditContributions.tsx
dominik-stumpf: The presentation part will be implemented here, presumably tomorrow.

src/app/(marketing)/profile/_components/EditContributions.tsx
dominik-stumpf: If I made it any smaller I think it would be too much. I will talk about this with the designers.

src/app/(marketing)/profile/_components/CircularProgressBar.tsx
```diff
@@ -0,0 +1,36 @@
+"use client"
+
+import { ChartContainer } from "@/components/ui/Chart"
+import { RadialBar, RadialBarChart } from "recharts"
```
dovalid, line 4: We should use visx (that we already use, we can talk about the reasons tomorrow) or ideally not even that just a simple Progress component with a basic solution (svg morphing maybe, but we should check what's the recommended way to do it)

src/app/(marketing)/profile/_components/ActivityChart.tsx
```diff
@@ -0,0 +1,100 @@
+"use client"
+
+import {
+  ChartConfig,
+  ChartContainer,
+  ChartTooltip,
+  ChartTooltipContent,
+} from "@/components/ui/Chart"
+import { Bar, BarChart, XAxis } from "recharts"
```
dovalid, line 9: Same here, we should use visx, but we should just remove the component now, we'll handle it when it's actually used

dominik-stumpf:
Merging this to remove the number of stale branches. Note that some of the changes here are not yet reviewed.

PR #1407, opened 2024-07-23
Playwright

BrickheadJohnny:
Initialized Playwright, added a simple test with the auth setup flow, and created a GitHub Action.


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,30 @@
...
+    # && !contains(github.event.deployment_status.target_url, 'guild-xyz-storybook')
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+      - name: Debug
+        run: echo "${{ github.event.deployment_status.target_url }}"
```
cs-balazs, line 16: I believe we don't need this not that the workflow works :ayy:

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Install dependencies
+        run: npm ci
```
cs-balazs, line 18: I'm wondering if we need to install all of the dependencies 🤔 Later on this might be worth an investigation

cs-balazs (approved):
Amazing 💯

PR #1411, opened 2024-07-24
feat: add container to StartProfile

dominik-stumpf:


next.config.js
```diff
@@ -88,6 +88,9 @@ const nextConfig = {
       {
         hostname: "og.link3.to",
       },
+      {
```
BrickheadJohnny, line 91: why do we need this? 👀

src/app/(marketing)/create-profile/_components/StartProfile.tsx
```diff
@@ -9,119 +11,211 @@ import {
...
     },
     mode: "onTouched",
   })
 
-  function onSubmit(values: z.infer<typeof formSchema>) {
-    console.log(values)
+  const createProfile = useCreateProfile()
+  async function onSubmit(values: z.infer<typeof profileSchema>) {
+    createProfile.onSubmit(values)
+    console.log("onSubmit", values)
```
BrickheadJohnny, line 65: We can remove this `console.log`

src/app/(marketing)/create-profile/_components/StartProfile.tsx
```diff
@@ -9,119 +11,211 @@ import {
...
+        description: error,
+      })
+    },
+  })
+
+  const [uploadProgress, setUploadProgress] = useState(0)
+  const { isDragActive, fileRejections, getRootProps } = useDropzone({
+    multiple: false,
+    noClick: false,
+    maxSizeMb: 4,
```
BrickheadJohnny, line 84: Let's just leave it on the default file size (afaik it's 5MB, but it'll fallback to that value if you don't specify the `maxSizeMb` prop). I'm pretty sure we'll get a feedback to increase it later. :D

src/app/(marketing)/create-profile/_components/StartProfile.tsx
```diff
@@ -9,119 +11,211 @@ import {
...
+    onDrop: (acceptedFiles) => {
+      if (!acceptedFiles[0]) return
+      onUpload({
+        data: [acceptedFiles[0]],
+        onProgress: setUploadProgress,
+      })
+    },
+  })
+
+  useEffect(() => {
```
BrickheadJohnny, line 94: Could we use `useDropZone`'s `onError` instead? That seems a bit cleaner to me.

src/app/(marketing)/create-profile/_components/StartProfile.tsx
```diff
@@ -9,119 +11,211 @@ import {
...
-          </ConnectFarcasterButton>
 
-          <Button variant="ghost">
-            I don't have a Farcaster profile
-            <ArrowRight weight="bold" />
-          </Button>
-        </>
-      )}
+              <Button
+                variant="ghost"
```
BrickheadJohnny, line 171: This could be a "link" variant

src/v2/components/Confetti.tsx
```diff
@@ -0,0 +1,101 @@
...
+        audioRef.current?.play()
+        return null
+      }
+      return confetti()
+    }
+    confettiClosure.reset = confetti.reset
+    confettiRef.current = confettiClosure
+  }
+  return (
+    <ConfettiContext.Provider value={confettiRef}>
```
BrickheadJohnny, line 88: I think we could provide a simple `playConfetti()` function instead of `confettiRef` here which we could simply call anywhere, so we wouldn't need to call `confetti.current?.()` (with optional chaining)

src/v2/components/Confetti.tsx
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export const ConfettiProvider = ({ children }: PropsWithChildren) => {
+  const confettiRef = useRef<TCanvasConfettiInstance>()
+  const audioRef = useRef<HTMLAudioElement>(null)
+  const onInitHandler = ({ confetti }: { confetti: TCanvasConfettiInstance }) => {
+    const confettiClosure = async (...args: Parameters<typeof confetti>) => {
+      if (args[0] === undefined) {
+        doubleConfetti(confetti)
+        audioRef.current?.play()
```
BrickheadJohnny, line 79: The sound is a bit much to me, idk what's your opinion on that hehe :D Maybe we should get feedback on that from the rest of the team!

BrickheadJohnny (commented):
Left some feedback ;)
Also, please make sure to handle the case when an unauthenticated user visits the page. I think we can display a connect wallet button (on step 1, instead of the Farcaster-related buttons) until the user is not connected to Guild. You can see a similar approach on the `/create-guild` page.

next.config.js
dominik-stumpf: This is the image provider from Farcaster.

src/app/(marketing)/create-profile/_components/StartProfile.tsx
```diff
@@ -9,119 +11,211 @@ import {
...
+        toast({
+          variant: "error",
+          title: `Failed to upload file "${file.name}"`,
+          description: error.message,
+        })
+      }
+    }
+  }, [fileRejections])
+
+  let avatarFallBackIcon = <User size={32} />
```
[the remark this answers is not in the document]
dominik-stumpf: Yes the conditionals are quite long and would look a bit messy in the markdown.

src/v2/components/Confetti.tsx
dominik-stumpf: Sure, for now I'm merging the sound, and remove it later, if needed.

src/v2/components/Confetti.tsx
dominik-stumpf: The optional chaining signifies that the confetti isn't always available (it takes time to load), but I will proceed with your suggestion.

src/app/(marketing)/create-profile/_components/StartProfile.tsx
dominik-stumpf: There is no such variant yet. How should it look like? @BrickheadJohnny

dominik-stumpf:
> Oh, and one more thing I noticed after the review: the image uploader doesn't trigger the file picker on click. We should fix that!

It works for me. The issue is maybe because there is no input for the dropzone?

src/app/(marketing)/create-profile/_components/StartProfile.tsx
BrickheadJohnny: It should look like an anchor component actually

BrickheadJohnny:
> > Oh, and one more thing I noticed after the review: the image uploader doesn't trigger the file picker on click. We should fix that!
> 
> It works for me. The issue is maybe because there is no input for the dropzone?
> 
> Merging for now because this is needed elsewhere. We should talk about the remaining issues later!

Okay!
Let's don't forget to remove the "Fill using Farcaster" button on the `profiles-page` branch, I think we don't need that anymore (and I think we can remove the `method` state too)! ;)

PR #1413, opened 2024-07-26
feat: Guild Pin minting E2E test

BrickheadJohnny:
Running the tests agains a Sepolia fork (Anvil) so we can actually run the onchain transactions and write more reliable tests 😎 

I also managed to find a way to use the built-in wagmi mock connector with a viem test client, so I removed our custom solution. :feelsgoodman:


src/v2/hooks/useAutoReconnect.ts
```diff
@@ -66,7 +70,7 @@ const useAutoReconnect = () => {
       connections: new Map(prevState.connections ?? []).set(
         connectorToReconnect.uid,
         {
-          accounts: data.accounts,
+          accounts: data.accounts as readonly [`0x{string}`, ...`0x${string}`[]],
```
cs-balazs, line 73: :nice:

cs-balazs (approved):
Amazing 💯

PR #1414, opened 2024-07-26
resolve circular dependencies

dominik-stumpf:
This PR intends to fix all remaining circular imports

Note: On the first commit, the loop is broken by removed module, which could cause issue in the current requirement creation flow, although it doesn't break anything. This was the easiest way of doing this as the `SelectRolePanel` requires all components.


src/hooks/useSubmit/fuelSign.ts
BrickheadJohnny, line 1: Couldn't we just put it to `src/hooks/useSubmit/utils.ts`? 👀

src/requirements/Form/components/AddFormButton.tsx
```diff
@@ -74,19 +62,6 @@ const AddFormButton = ({ baseFieldPath, isDisabled }) => {
           }}
         />
       )}
-
-      {step === "SELECT_ROLE" && (
```
BrickheadJohnny, line 78: Unfortunately if the user won't add the form to a role as a reward, they won't be able to use it as a requirement neither. I know, this is a pretty strange behaviour, but this change isn't correct because of that :/

dominik-stumpf: Indeed. I have an alternative solution here we should talk about. :wink:

BrickheadJohnny (approved):
🚀

PR #1419, opened 2024-07-29
fix(EditGuildDrawer): validate `eventSources` with Zod

BrickheadJohnny:
Important: this is a hotfix. We'll need to write a proper schema for this form in our types package & use Zod for validating every field in the edit guild form.


dominik-stumpf (approved):
:rocket:

PR #1423, opened 2024-07-31
feat: rework the `Layout` component

BrickheadJohnny:
Changes:
- Every sub-component can handle the `className` (and other props) which we pass to it
- The `LayoutHeadline` component only handles the flex layout, I extracted the `LayoutTitle` component so we can use it separately (we'll need this kind of behaviour on the `/:guildId` page)
- Removed unnecessary nesting in `LayoutMain` & `LayoutFooter`

Cleanup:
- Removed comment blocks
- Removed the `Header` re-export to avoid confusion


src/v2/components/Layout/Layout.tsx
```diff
@@ -1,125 +1,98 @@
...
-
-const LayoutHeadline = ({ title }: LayoutHeadlineProps) => (
-  <LayoutContainer>
-    <h1 className="pt-9 pb-14 font-bold font-display text-4xl text-white tracking-tight sm:text-5xl">
-      {title}
-    </h1>
-  </LayoutContainer>
+const LayoutBanner = ({ className, ...props }: HTMLAttributes<HTMLDivElement>) => (
+  <div
+    className={cn("-z-10 absolute inset-0 overflow-hidden bg-banner", className)}
```
dominik-stumpf, line 46: The banner is meant to be an element defining the area of its background, since not every page requires a `bg-banner`. Should we overwrite the value on the fly?

dominik-stumpf (commented):
Good job with the refactor, what do you think about including each variant of the Layout used in action in a storybook? For example the incoming `/profile` variant, `/create-profile` and the new `/explorer`.

src/v2/components/Layout/Layout.stories.tsx
```diff
@@ -7,24 +8,36 @@ import {
...
 
 export const Static: StoryFn = () => (
   <Layout>
     <LayoutHero>
-      <LayoutBanner>
-        <LayoutBannerBackground />
-      </LayoutBanner>
-      <LayoutHeadline title="Layout title" />
+      <LayoutBanner />
+      <header className="relative flex h-16 w-full items-center justify-between p-2">
```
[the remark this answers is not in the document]
dovalid: I think it's cool to delete, this doesn't provide much value in storybook

src/v2/components/Layout/Layout.tsx
BrickheadJohnny: Yup, that was my idea. Providing a good looking default value here which we can overwrite when needed.

PR #1428, opened 2024-08-02
feat(guild page): remove members section & redesign cover

BrickheadJohnny:
- Started using the new `Layout` component
- Redesigned the cover section (the image got bigger, so we can put the member count badge below the guild name
- Removed the members section

In the future, I'd like to add a `--banner-foreground` variable & control it from `ThemeContext`, I'm pretty sure that would simplify the color management in the cover section a lot.


src/pages/[guild]/index.tsx
```diff
@@ -122,126 +122,173 @@ const GuildPage = (): JSX.Element => {
...
+                        textColor === "primary.800",
+                    }
+                  )}
+                >
+                  {name}
+                </LayoutTitle>
+                <CheckMark className="mt-2.5 size-5 shrink-0 sm:mt-4 sm:size-6" />
+              </div>
+
+              <Badge
```
Comment (dovalid) [inline src/pages/[guild]/index.tsx:196]: I think we should use the common MemberCount component here. I know that's not Tailwind yet, we'll be able to migrate it there once and it'll change everywhere

PR #1431, opened 2024-08-06
feat: remove tabs & add `/dashboard` page

BrickheadJohnny:


src/components/common/Layout/components/BackButton.tsx
```diff
@@ -7,15 +7,16 @@ import { PropsWithChildren } from "react"
 
 type Props = {
   href?: string
+  forceRender?: boolean
```
dovalid, line 10: Instead of this new prop, we should separate the component into a general BackButton and a BackToExplorer button, and use the latter where we used BackButton so far, and use the simple BackButton for the new dashboard navigation

src/pages/[guild]/dashboard/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+import Card from "components/common/Card"
+import Section from "components/common/Section"
+import Head from "next/head"
+
+const DashboardPage = () => {
+  const { isDetailed, name } = useGuild()
+  const { isOwner } = useGuildPermission()
+  const { isSuperAdmin } = useUser()
+
+  if (!isDetailed)
```
Comment (dovalid) [inline src/pages/[guild]/dashboard/index.tsx:29]: Would be great to handle with NoPermissionToPageFallback if I open the link but don't have proper auth, now we get an infinite loading state in that case

PR #1433, opened 2024-08-07
chore: adjust v2 ui components for EditContribution

dominik-stumpf:


src/v2/components/ui/Select.tsx
dovalid, line 1: Would be great to have a storybook for this. Other than that, consider the PR approved!

dovalid (approved):

PR #1435, opened 2024-08-08
cleanup: remove PolygonID reward

BrickheadJohnny:
TODO before merge: remove PolygonID rewards from our DB


BrickheadJohnny:
We actually won't remove this reward, we'll rework it instead

PR #1436, opened 2024-08-08
feat: simplify access hub

BrickheadJohnny:


src/components/[guild]/RoleCard/components/RoleRewards.tsx
```diff
@@ -0,0 +1,73 @@
...
+          </SlideFade>
+        )
+      })}
+
+      {role.hiddenRewards && <HiddenRewards />}
+    </div>
+  )
+}
+
+const RewardWrapper = ({
```
Comment (dovalid) [inline src/components/[guild]/RoleCard/components/RoleRewards.tsx:59]: Why do we need this, can't we handle the provider in ClaimTokenButton?

src/components/[guild]/AccessHub/AccessHub.tsx
```diff
@@ -31,128 +15,47 @@ const DynamicCreatedPageCard = dynamic(() =>
...
     ((isMember && guildPin?.isActive) || isAdmin)
 
   const hasVisiblePages = !!groups?.length && roles?.some((role) => !!role.groupId)
 
-  const showAccessHub =
-    isAdmin ||
-    isMember ||
-    !!accessedGuildPlatforms?.length ||
-    (hasVisiblePages && !group)
+  const showAccessHub = isAdmin || isMember || (hasVisiblePages && !group)
```
Comment (dovalid) [inline src/components/[guild]/AccessHub/AccessHub.tsx:32]: Shouldn't we remove the isMember condition? We should show it if there are visible pages now, or for the admin to set up pages / pin

Comment (BrickheadJohnny) [reply src/components/[guild]/AccessHub/AccessHub.tsx]: You're right, and we should actually add the `shouldShowGuildPin` boolean, fixed in https://github.com/guildxyz/guild.xyz/pull/1436/commits/a00fb20d4d7cf41fde49cea195a3ccff37d4e766!

src/components/[guild]/RoleCard/components/RoleRewards.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/RoleCard/components/RoleRewards.tsx]: Nice catch, I honestly don't remember why I did this, and your suggestion will work just fine, so refactored it in https://github.com/guildxyz/guild.xyz/pull/1436/commits/1190bde520720c333ffcbf6f02cc9439c3ec861a.

PR #1439, opened 2024-08-12
implement activity log container for `/profile`

dominik-stumpf:
Implements activity log fetching and filtering using `/v2/audit-log` auth endpoint.


dominik-stumpf:
We decided to use the existing endpoint for fetching activities and reuse existing code which will be in another PR.

PR #1440, opened 2024-08-12
chore: add empty contribution card

dominik-stumpf:


src/app/(marketing)/profile/_components/EditContributions.tsx
```diff
@@ -117,6 +117,14 @@ export const EditContributions = () => {
         </DialogHeader>
         <DialogBody className="gap-7">
           <div className="flex flex-col gap-3">
+            {guilds?.length === 0 && (
```
dovalid, line 120: We should abstract the "Contributions will appear here" card into a ContributionsFallback component, show that here too in case of no contributions, and handle the no verified guilds state in the select (disable it and show info as FormHelpMessage). Related: we should also prevent the selection of guilds of already added contributions - either filter them from all guild options and so the select could be disabled because of that too, or make the options themeselves disabled, whichever is easier

dominik-stumpf:
Relevant changes copied to #1447, closing for that reason.

PR #1441, opened 2024-08-12
feat: add models to `/create-profile` scenes

dominik-stumpf:
This PR adds 3D `.glb` models to guild pass scene for the `/create-profile` process.


dominik-stumpf:
Not final version of the scene. Merging to reduce the number of stale branches.

PR #1444, opened 2024-08-14
feat: migrate the `/create-guild` page to Tailwind

BrickheadJohnny:
Additional changes:
- using Zod for form validation
- updated the input styles to match the Chakra UI inputs
- simplified the icon picker, now it's just a simple image uploader (this was a request from the product team)


src/app/create-guild/_components/CreateGuildImageUploader.tsx
```diff
@@ -0,0 +1,119 @@
...
+import usePinata from "hooks/usePinata"
+import { useState } from "react"
+import { useFormContext, useWatch } from "react-hook-form"
+import { CreateGuildFormType } from "../types"
+
+const MIN_WIDTH = 512
+const MIN_HEIGHT = 512
+
+/**
+ * This is a pretty specific component right now, but we should generalise it once we start using it in other places too (e.g. on the create profile page)
```
BrickheadJohnny, line 18: I wouldn't like to implement unnecessary logic, which we possibly won't actually use in the future, that's why I decided to go with a "specific" component here. ;)

dovalid (commented):
Awesome! Left some adjustments, now the only problem is that the image uploader doesn't seem to be working for me. I know you'll be OOO so I'll handle and merge it, just noting here :ayy:

PR #1445, opened 2024-08-14
Remake `/create-profile` driver with referrer

dominik-stumpf:
Implementation goals:

- Create referrer feature + validation for claim pass
- Create client side authentication wall: redirect unauthenticated users.
- Redirect restriction logic between chain steps, e.g. redirect from start profile if referred user or subscription is not present
- Make funnel navigation possible using history API

> Note: 
> Username input in claim pass is not debounced yet as it is planned to be addressed in another PR.
> Initially, we intended to use the `userId` for the referrer profile, but this approach led to backend difficulties. As a result, we decided to switch to using the `username` refers instead.
> For faster iterations we are temporarily moving on to dev API.


src/app/(marketing)/create-profile/(onboarding)/types.ts
```diff
@@ -1,14 +1,19 @@
+import { Schemas } from "@guildxyz/types"
 import { FunctionComponent } from "react"
 import { SUBSCRIPTIONS } from "./constants"
 
 export type ChainAction = "next" | "previous"
+
 export interface ChainData {
```
dovalid, line 7: Why do we call these `chain`? I'd except CreateProfileData or something like that

src/app/(marketing)/create-profile/(onboarding)/_components/ChainSkeleton.tsx
```diff
@@ -0,0 +1,5 @@
+import { Skeleton } from "@/components/ui/Skeleton"
+
+export const ChainSkeleton = () => {
```
dovalid, line 3: (same here)

PR #1446, opened 2024-08-16
Exclude subscription features from `/create-profile`

dominik-stumpf:
This reverts commit 56e64e8b29c465e65123449b654b7ea0e31c0375.


dominik-stumpf:
We decided to remove the subscription feature altogether and extract it that way, so merging this back.

PR #1455, opened 2024-08-28
Add farcaster follower and relevant follower integration to `/profile`

dominik-stumpf:
### Objective

Using [neynar API](https://docs.neynar.com/reference/followers-v2) retrieve followers, relevant (common) followers and followings and display them in `ProfileHero`

### Implementation details

Failed to find endpoint to limit and retrieve the number of relevant followers (we could use length on the response, but there isn't enough data to work with), more testing is needed to see how relevant followers feel like. For now it is exchanged with regular follower count depending on auth state.

Reference: 
- https://docs.neynar.com/reference/relevant-followers
- https://docs.neynar.com/reference/followers-v2


dominik-stumpf:
Should we place a detailed view of FC data and link to FC profile?

dovalid (approved):

dovalid:
I ended up separating the two hooks and moved them to a separate file, it only felt declarative this way. Now I feel like everything is nice, self-explaining and modular. Also adjusted the SWR calls to use them as designed, so it's consistent and we have loading states etc. Approved the PR, please review the changes, and merge if alls right!

dominik-stumpf:
Great work! Thanks for the help and the cleanup. :rocket:

PR #1456, opened 2024-08-28
feat: add reorder to requirements [GUILD-3711]

dominik-stumpf:
### Objective

Make reordering possible in the `EditRequirements` section of `EditRole` drawer, similar to how role reordering is currently implemented.

Adapt code from `OrderRolesModal` and it's submodules.

### Implementation details

The role order after reordering is sent as `PUT` through the `/v2/guilds/{{guildId}}/roles/{{roleId}}` endpoint with an additional `position` value in the payload. The same is implemented for requirement, but with the `/v2/guilds/{{guildId}}/roles/{{roleId}}/requirements/{{requirementId}}` endpoint to change ordering.


dominik-stumpf:
Stale branch, but a feature to keep in mind. Not relevant changes anymore due to V3 migration, closing.

PR #1457, opened 2024-08-28
feat: add rich text support for requirements

dominik-stumpf:
### Objective

Add rich text support (utilizing a restricted set markdown and GFM) to custom requirement paragraph.

### Motivation

By itself the pre-written requirements are well formatted and there is an option to create custom ones. However they don't support markdown or any rich text as some users would expect/desire.

### Considerations

- Backward compatibility
- Theme fit
- Use tailwind styles and plugins
- Disallow inappropriate elements

> [!NOTE]
> Allowed HTML tags are: `a`, `p`, `code`, `del`, `strong`, `em`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "react-markdown": "^8.0.7",
     "react-papaparse": "^4.4.0",
     "react-window": "^1.8.8",
-    "remark-gfm": "^4.0.0",
+    "remark-gfm": "^3.0.1",
     "slugify": "^1.6.5",
     "swr": "^2.2.4",
     "tailwind-merge": "^2.3.0",
--- a/src/components/[guild]/Requirements/components/Requirement.tsx
+++ b/src/components/[guild]/Requirements/components/Requirement.tsx
@@ -12,6 +12,8 @@ import Visibility from "components/[guild]/Visibility"
 import dynamic from "next/dynamic"
 import React, { ComponentType, PropsWithChildren } from "react"
 import { useFormContext } from "react-hook-form"
+import ReactMarkdown from "react-markdown"
+import remarkGfm from "remark-gfm"
 import { useRequirementContext } from "./RequirementContext"
 import { RequirementImage, RequirementImageCircle } from "./RequirementImage"
 import ResetRequirementButton from "./ResetRequirementButton"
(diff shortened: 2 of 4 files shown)


tailwind.config.ts
```diff
@@ -206,9 +222,110 @@ const config = {
         modal: "1400",
         popover: "1500",
         toast: "1700",
-        tooltip: "1800", 
+        tooltip: "1800",
       },
     },
+    typography: () => ({
```
dovalid, line 228: Why do we need these custom overrides?

package.json
```diff
@@ -104,7 +104,7 @@
     "react-markdown": "^8.0.7",
     "react-papaparse": "^4.4.0",
     "react-window": "^1.8.8",
-    "remark-gfm": "^4.0.0",
+    "remark-gfm": "^3.0.1",
```
dovalid, line 107: Why do we have to downgrade the package?

src/components/[guild]/Requirements/components/Requirement.tsx
```diff
@@ -71,9 +73,21 @@ const Requirement = ({
       <VStack alignItems={"flex-start"} alignSelf="center" spacing={1.5}>
         <ChildrenWrapper display="inline-block">
           {requirement?.isNegated && <Tag mr="2">DON'T</Tag>}
-          {requirement?.type === "LINK_VISIT"
-            ? children
-            : requirement?.data?.customName || children}
+          {requirement?.type === "LINK_VISIT" ? (
```
Comment (dovalid) [inline src/components/[guild]/Requirements/components/Requirement.tsx:76]: I feel like it could be confusing that we don't support markdown for link requirements, while we support it for everything else. Ideally we should enable it for that too, handling the markdown rendering in VisitLinkRequirement too (so the necessary custom Link rendering remains too, meaning this condition will still stay here). Keeping our original syntax that in theory works well with markdown, so a [] without a following (url) transforming to the clickable link to visit, and everyhing else automatically by the markdown. But if there's any complication that comes up regarding this, don't spend a lot of time with it, it's ok to go with the unsupport for now then

dovalid (commented):
Exciting feature, left my thoughts for later when we've launched profile!

dominik-stumpf:
Stale branch. Not relevant changes anymore due to V3 migration, where this feature would be built-in. Closing.

PR #1459, opened 2024-09-02
fix(useCreateRequirement(s)): mutate `requiredPlatforms`

BrickheadJohnny:
When an admin added a new requirement which required a platform connection (e.g. Twitter, Farcaster, Email, etc.), and then opened the join modal (with another account, without refreshing the page), the newly added `requiredPlatform`s weren't included in the `/guild-page` SWR cache. This PR aims to fix this issue.

My first idea was to just add an if statement inside the `useCreateRequirement` and `useCreateRequirements` hooks, but I'm 100% sure I'd forgot to update these conditions when we add a new platform type to guild, so that's why I went with this solution instead.

### DIFF
--- a/src/components/[guild]/AddRewardButton/hooks/useCreateRequirements.ts
+++ b/src/components/[guild]/AddRewardButton/hooks/useCreateRequirements.ts
@@ -8,10 +8,11 @@ import {
 } from "types"
 
 import { useFetcherWithSign } from "hooks/useFetcherWithSign"
+import { JOINABLE_REQUIREMENT_PLATFORMS } from "rewards"
 import preprocessRequirement from "utils/preprocessRequirement"
 
 const useCreateRequirements = () => {
-  const { id: guildId } = useGuild()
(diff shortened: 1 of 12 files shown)


dovalid (changes requested):
I feel like we can't justify this much custom logic for such an edge case. We should just make sure to refetch the guild endpoint when an admin switches account - that way we can be sure that everything is fresh and is according to the new account's privileges, we can't have custom mutations for such cases cause it's really complex and I'm sure this is not the only case to handle. Wdyt does that solve the problem?

BrickheadJohnny:
> I feel like we can't justify this much custom logic for such an edge case. We should just make sure to refetch the guild endpoint when an admin switches account - that way we can be sure that everything is fresh and is according to the new account's privileges, we can't have custom mutations for such cases cause it's really complex and I'm sure this is not the only case to handle. Wdyt does that solve the problem?

Hmm, yep, that would be a better solutions & it would probably solve other similar edge cases too. I'll think about a solution this week!

PR #1461, opened 2024-09-04
feat: use unique SWR keys for authenticated API requests

BrickheadJohnny:
The keys for these requests weren't unique, so users (and especially guild admins) could encounter strange issues. This PR aims to solve these issues by adding the user ID to the SWR keys.

**How to test?**

Before the fix:
1. Open the explorer page
2. Log in *(user A)*
3. Wait until the guilds in the "Your guilds" section are fetched
4. Log out
5. Log in to another account *(user B)*
6. You'll see the guilds of *user A*

After the fix:
You'll always see the correct guilds on the explorer page.


cs-balazs (approved):
We discussed some questions IRL, thanks for the changes 🚀

PR #1463, opened 2024-09-04
feat: allow editing point rewards

BrickheadJohnny:


dovalid (changes requested):
Nice, it works amazing for static points! For dynamic points tho, the base value is not shown. Also, switching between static and dynamic causes issues, cause they're stored differently and we'd have to manually handle removing the other key etc.. I wouldn't walk into that street, just disable the static / dynamic switcher when editing, and look into the default base value in case of dynamic, that should be an easy fix. That way we'll be able to edit both static and dynamic points correctly, just not switch between them, which is awesome

BrickheadJohnny:
> Nice, it works amazing for static points! For dynamic points tho, the base value appears to be not shown. Also, switching between static and dynamic causes issues, cause they're stored differently and we'd have to manually handle removing the other key etc.. I wouldn't walk into that street, just disable the static / dynamic switcher when editing, and look into the default base value in case of dynamic, that should be an easy fix. That way we'll be able to edit both static and dynamic points correctly, just not switch between them, which is awesome

Done & done 👍 We've talked about the base value issue on Discord, for now I wrote a simple mapper function as a "hotfix", but also left a comment about a better solution for the future! 😉

PR #1464, opened 2024-09-04
feat: migrate the `/oauth` & `/oauth-result` pages to app router

BrickheadJohnny:
The PR is 90% copy-paste, the only thing that changed is that we can now use `searchParams` from page props instead of the `useRouter` hook & of course we use Tailwind CSS for the UI.


cs-balazs (approved):
Nice, lgtm 🚀 
I'm wondering if we can test this out somehow 🤔

BrickheadJohnny:
> I'm wondering if we can test this out somehow 🤔

+1, let's talk about it IRL later today 😉

PR #1465, opened 2024-09-05
feat: allow appending dash "-" to the end of url name

dominik-stumpf:
### Objective

Allow appending dash `-` character to the end of the url for convenience

Linear issue: [GUILD-3545](https://linear.app/guildxyz/issue/GUILD-3545/guild-url-name-accept-too)

> Note: I don't feel like it's a bad limitation, but `slugify` doesn't allow neighbouring dashes.

### DIFF
--- a/src/components/[guild]/EditGuild/components/UrlName.tsx
+++ b/src/components/[guild]/EditGuild/components/UrlName.tsx
@@ -18,8 +18,7 @@ const checkUrlName = (urlName: string) =>
 
 const UrlName = () => {
   const { errors } = useFormState()
-  const { register, setError, clearErrors, setValue } = useFormContext()
-
+  const { register, setError, clearErrors, getFieldState } = useFormContext()
   const { urlName: currentUrlName } = useGuild()
 
   return (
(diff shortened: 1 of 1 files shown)


src/components/[guild]/EditGuild/components/UrlName.tsx
```diff
@@ -30,9 +30,14 @@ const UrlName = () => {
         <Input
           {...register("urlName")}
           onChange={(event) => {
-            setValue("urlName", slugify(event.target.value), {
-              shouldDirty: true,
-            })
+            setValue(
```
Comment (BrickheadJohnny) [inline src/components/[guild]/EditGuild/components/UrlName.tsx:33]: What if we'd just define a custom validator inside the `register` function which checks if `value === slugify(value)` and we display an "Invalid URL name" error message otherwise? 👀 
That way we could get rid of the custom `onChange` handler and we wouldn't modify the user input on-the-fly. This way the user would be able to write a dash, but they wouldn't be able to submit the form, since it would be invalid that way. Wdyt?

Comment (dominik-stumpf) [reply src/components/[guild]/EditGuild/components/UrlName.tsx:33]: Sure thing! Using validation is much more gentle on the user than overwriting the input value. 😉

BrickheadJohnny (approved):
🚀

PR #1467, opened 2024-09-05
improve ui `Button` loading state

dominik-stumpf:
### Motivation

Currently our loading state without loading text shows an empty spinner with an odd space on the right (due to flexbox gap on `Button`).

### Resolution

If `loadingText` is missing, we should fallback to default `children` prop.

- Added `loadingText` to storybook as a bonus! :star:

### DIFF
--- a/src/v2/components/ui/Button.stories.ts
+++ b/src/v2/components/ui/Button.stories.ts
@@ -17,6 +17,7 @@ export const Solid: Story = {
     size: "md",
     variant: "solid",
     colorScheme: "primary",
+    loadingText: undefined,
     isLoading: false,
     disabled: false,
   },
@@ -25,6 +26,10 @@ export const Solid: Story = {
       type: "string",
       control: "select",
       options: ["xs", "sm", "md", "lg", "xl"] satisfies ButtonProps["size"][],
+    },
+    loadingText: {
+      type: "string",
+      control: "text",
     },
     colorScheme: {
       type: "string",
--- a/src/v2/components/ui/Button.stories.tsx
+++ b/src/v2/components/ui/Button.stories.tsx
@@ -1,5 +1,5 @@
+import { Book } from "@phosphor-icons/react"
 import type { Meta, StoryObj } from "@storybook/react"
-
 import { Button, ButtonProps } from "./Button"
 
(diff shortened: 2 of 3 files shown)


BrickheadJohnny (approved):
🚀

BrickheadJohnny (changes requested):
I approved the PR, then I realised that this unfortunately won't work properly for buttons with icons on them (we'd render a loading spinner, and icon, and a text on the button):
image

Let's just don't render anything next to the spinner if `loadingText` is not defined. We should think about a better solution later!

dominik-stumpf:
> Let's just don't render anything next to the spinner if `loadingText` is not defined. We should think about a better solution later! 

I could extract the text from the JSX element and the discord icon would be replaced. What do you think? @BrickheadJohnny

BrickheadJohnny:
> > Let's just don't render anything next to the spinner if `loadingText` is not defined. We should think about a better solution later!
> 
> I could extract the text from the JSX element and the discord icon would be replaced. What do you think? @BrickheadJohnny

Would that be a hacky solution? :D I'd go with the simplest solution for now, since we'll probably rework our Button component soon anyways.

dominik-stumpf:
Now this works well, but It's kind of a hacky solution... We could also use an `Icon` prop on the `Button` and replace that with the spinner or create a separate component `LoadingButton` or something else. Should we brainstorm on it?

BrickheadJohnny:
> Now this works well, but It's kind of a hacky solution... We could also use an `Icon` prop on the `Button` and replace that with the spinner or create a separate component `LoadingButton` or something else. Should we brainstorm on it?

Ehh, yep, let's go with the simple solution which I mentioned above for now. We'll think about this issue next week in the office! 😉

BrickheadJohnny:
I just started working on an alternative solution - `leftIcon` & `rightIcon` props. This way we can easily replace the left icon in loading state, and we also won't have the unnecessary gap in case we don't provide a loading text.
We like this API in Chakra UI, so why wouldn't we keep using it this way? 😉 

Check here: https://github.com/guildxyz/guild.xyz/pull/1492

BrickheadJohnny:
Merged the other PR, so closing this one!

PR #1474, opened 2024-09-10
feat: `DiscordCardWarning` component

BrickheadJohnny:
Added a `cardWarningComponent` for Discord rewards which indicates if the original Discord role doesn't exist anymore.

Also improved the code a bit:
- Removed the custom logic for `GoogleCardWarning` from `Reward.tsx`, we don't need that anymore
- Cleaned up the `GoogleCardWarning` component
- Removed empty import lines
- Created a reusable `CardWarningComponentBase` component

Note:
We should refactor the fetching logic in the `DiscordBotPermissionsChecker` component, I noticed that it won't write in the `/v2/discord/servers/:serverId/permissions` cache, so we'll actually fetch the data 2 times in some edge cases. This is a low prio task & is not the scope of this PR, so I left it as is for now.

### DIFF
--- a/src/components/[guild]/AccessHub/components/AccessedGuildPlatformCard.tsx
+++ b/src/components/[guild]/AccessHub/components/AccessedGuildPlatformCard.tsx
@@ -33,9 +33,7 @@ const AccessedGuildPlatformCard = () => {
       guildPlatform={rolePlatform.guildPlatform}
       cornerButton={
         <>
-          {PlatformCardWarning && (
-            <PlatformCardWarning rolePlatform={rolePlatform} />
-          )}
+          {PlatformCardWarning && <PlatformCardWarning />}
           {isAdmin && isDetailed && PlatformCardMenu && (
             <PlatformCardMenu
               platformGuildId={rolePlatform.guildPlatform.platformGuildId}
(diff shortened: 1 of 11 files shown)


src/components/[guild]/AccessHub/components/AccessedGuildPlatformCard.tsx
```diff
@@ -2,45 +2,53 @@ import PlatformCard from "components/[guild]/RolePlatforms/components/PlatformCa
...
-import rewards from "rewards"
 import { cardPropsHooks } from "rewards/cardPropsHooks"
-import rewardComponents from "rewards/components"
+import rewardComponentsConfig from "rewards/components"
 import { PlatformName, PlatformType } from "types"
 import PlatformAccessButton from "./PlatformAccessButton"
 
 const AccessedGuildPlatformCard = () => {
-  const { guildPlatform } = useRolePlatform()
+  const rolePlatform = useRolePlatform()
```
Comment (dovalid) [inline src/components/[guild]/AccessHub/components/AccessedGuildPlatformCard.tsx:11]: Nit: why not keep the {guildPlatform} destruction here? Since we only use that from rolePlatform, felt cleaner to me

dovalid (approved):

src/components/[guild]/AccessHub/components/AccessedGuildPlatformCard.tsx
Comment (BrickheadJohnny) [reply src/components/[guild]/AccessHub/components/AccessedGuildPlatformCard.tsx:11]: Oh, good point, I originally passed down `rolePlatform` as a prop to the `PlatformCardWarning` component, but then realised that I could just use the `useRolePlatform` hook & just forgot to revert this change. Will do it before merging the PR, thanks!

PR #1480, opened 2024-09-12
feat(forms): Edit responses

dovalid:


src/rewards/Forms/hooks/useFormSubmissions.tsx
```diff
@@ -89,7 +89,7 @@ const useFormSubmissions = (formId, queryString) => {
   return { data: flattenedData, ...rest }
 }
 
-const useUserFormSubmission = (form: Schemas["Form"]) => {
+const useUserFormSubmission = (form?: Schemas["Form"]) => {
```
BrickheadJohnny, line 92: nit: is there a reason this is optional now? 👀

src/components/[guild]/forms/FillForm.tsx
```diff
@@ -129,6 +138,27 @@ const FillForm = ({ form }: Props) => {
...
+
+  if (!userSubmission) return <FillForm form={form} />
+
+  if (form.isEditable)
+    return (
+      <FillForm
+        form={form}
+        method="PUT"
+        defaultValues={
+          userSubmission && mapAnswersToObject(userSubmission.submissionAnswers)
```
Comment (BrickheadJohnny) [inline src/components/[guild]/forms/FillForm.tsx:154]: I'd just put this inside the if condition above, just to make sure we don't ever render the edit form component without default values

src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddFormPanel/AddFormPanel.tsx
```diff
@@ -64,18 +64,23 @@ const AddFormPanel = ({ onAdd }: AddRewardPanelProps) => {
...
-            w="max-content"
-            ml="auto"
-            onClick={methods.handleSubmit(onSubmit, console.error)}
-            loadingText="Creating form"
-            isLoading={isLoading}
-            isDisabled={!fields?.length}
-          >
-            Create form & continue
-          </Button>
+          <Flex mt="4" justifyContent="space-between" alignItems="end">
```
Comment (BrickheadJohnny) [inline src/components/[guild]/RolePlatforms/components/AddRoleRewardModal/components/AddFormPanel/AddFormPanel.tsx:67]: I think we should display the text and button below each other on smaller screens, wdyt? Maybe the text should be below the button actually in that case.

src/rewards/Forms/FormCardMenu.tsx
```diff
@@ -29,14 +29,16 @@ const FormCardMenu = ({ platformGuildId }: Props): JSX.Element => {
...
-        <LinkMenuItem
-          href={`/${urlName}/forms/${formId}/responses`}
-          icon={<Table />}
-        >
-          View responses
-        </LinkMenuItem>
-
         <RemovePlatformMenuItem platformGuildId={platformGuildId} />
+
+        {form?.isEditable && (
```
BrickheadJohnny, line 34: Whoops, `!!userSubmission` is missing here, I guess? 👀

src/rewards/Forms/FormCardLinkButton.tsx
```diff
@@ -21,29 +21,48 @@ const FormCardLinkButton = ({ platform }: Props) => {
 
   const { userSubmission, isValidating } = useUserFormSubmission(form)
 
-  if (isAdmin && !!userSubmission)
+  if (isFormsValidating || isValidating) return <RewardCardButton isLoading />
+  if (!form) return <RewardCardButton isDisabled>Form not found</RewardCardButton>
+
+  if (!userSubmission)
```
BrickheadJohnny, line 27: I'm not sure if we should just move the `isAdmin` block before this one (that way the admin won't be able to fill the form easily?), or we should actually move the "view responses" link to the reward card menu. What's your opinion on this one?

BrickheadJohnny (commented):
Left a couple of comments, but I like the PR! 👌

src/rewards/Forms/hooks/useFormSubmissions.tsx
BrickheadJohnny: form can be sometimes undefined (because we get its value from a useSWR hook), that's why it's optional here

PR #1482, opened 2024-09-12
fix: regenerate package-lock

BrickheadJohnny:


.github/workflows/code-check.yml
```diff
@@ -10,10 +10,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Setup Biome
-        uses: biomejs/setup-biome@v2
       - name: Run Biome
-        run: biome ci --reporter=github .
+        run: npx biome ci --reporter=github .
```
dominik-stumpf, line 14: This doesn't seem to feed it's output to github properly, behind the scene that action does more than running `ci`. What went wrong that requires this change?

BrickheadJohnny: We don't know :D
The GH action couldn't run for some reason, but we couldn't see any error messages

dominik-stumpf: Interesting. It looks like the `ci` exits the moment error happens. Is there a local error, that could be triggering the      GH action?

dominik-stumpf: `npx biome ci --diagnostic-level=error` gave out a small formatting error, checking if that is the cause

BrickheadJohnny: Approved your PR, ty! Would be good to see this kind of errors on GitHub somehow tho :D

dominik-stumpf: Now it shows errors on my PR, but they are invalid? Something is fishy with this action. :fish:

BrickheadJohnny: 😵‍💫

PR #1483, opened 2024-09-12
fix biome error

dominik-stumpf:
- **style: apply biome write-ci**
- **revert: take back biome ci change**


BrickheadJohnny (approved):

BrickheadJohnny (approved):
Thank you! Tested it locally too and seems like it works great now!

PR #1488, opened 2024-09-14
cleanup: remove cwaas

BrickheadJohnny:
I left a comment in `useSubmit/utils.ts`, I wasn't sure if I can remove the related code there. Otherwise, I think the PR is pretty straightforward :D

And we should also remove the related stuff from Vercel, I guess.


src/hooks/useSubmit/utils.ts
```diff
@@ -87,7 +87,7 @@ export const sign = async ({
     params.chainId ||= chainId || `${walletClient.chain.id}`
 
     if (walletClient?.account?.type === "local") {
```
BrickheadJohnny, line 89: I see that we check this prop in `useSetKeyPair` & `useLinkAddress` too, but I wasn't sure if I can remove that code of we should keep it.

cs-balazs: I think these checks should be safe to remove 👍

cs-balazs (approved):

PR #1493, opened 2024-09-18
feat: `IconButton` component & stories

BrickheadJohnny:
Implemented this component for 2 reasons:
- We'll refactor the `Button` component & it'll receive left and right icon as a prop, so I'd like to maintain a consistent API for this (we could technically use the button component & pass the icon as a child to it too, but then we'd probably forget to define the `aria-label` attribute, which is important for accessibility)
- I noticed that we have an `icon` and an `icon-sm` size variant in our `Button` component, which in my opinion clearly indicated that we might need a new component for these variants


dominik-stumpf (approved):
Nice way of extending `Button` and the `aria-label` is a smart addition 😌

PR #1503, opened 2024-09-24
fix: handle loading wallet client

BrickheadJohnny:
Sometimes the wallet client is still `undefined` when the user clicks on the "Verify address" button, so the users gets on error. This PR aims to handle this edge case.


cs-balazs (approved):
Looks great! 🚀

PR #1507, opened 2024-09-25
feat: add download anchor to leaderboard endpoint

dominik-stumpf:
### Objective

Make it easier for guild admins to view and export their leaderboard data for personal use.

### Implementation details

~~Opted for redirect instead of file download, since it is more straightforward and allows to get live data on demand.~~

Create a dynamic route handler that converts the leaderboard `json` to `csv` for ease of use.

- [discord thread](https://discord.com/channels/697041998728659035/1080486578272620574/1288126534011457568)


src/components/[guild]/leaderboard/LeaderboardPointsSelector.tsx
```diff
@@ -30,9 +32,25 @@ const LeaderboardPointsSelector = () => {
   if (pointsRewards.length < 2) {
     if (isAdmin)
       return (
-        <Card borderRadius="xl" flexShrink={0}>
-          <RecalculateLeaderboardButton />
-        </Card>
+        <ButtonGroup>
```
Comment (BrickheadJohnny) [inline src/components/[guild]/leaderboard/LeaderboardPointsSelector.tsx:35]: This will only be displayed if there are less than 2 point rewards, which is not correct hehe. And while I like this new design, I'd go for consistency here & use the same design for all cases.

BrickheadJohnny (changes requested):

PR #1510, opened 2024-09-26
feat: add interactivity to profile

dominik-stumpf:


src/app/(marketing)/profile/_components/OperatedGuilds.tsx
```diff
@@ -85,10 +89,17 @@ export const OperatedGuild = ({ guildBase }: { guildBase: GuildBase }) => {
           <EmphasizedData>{rewardCount}</EmphasizedData>
           rewards in total
         </OperatedGuildDetail>
-        {createdAt && (
+        {createdAt && guild.createdAt && (
```
BrickheadJohnny, line 92: Why do we need both conditions here? 👀 `createdAt` already checks is `guild.createdAt` is defined.

BrickheadJohnny (approved):
Simplified the code a bit and left one comment, but LGTM overall!

PR #1518, opened 2024-10-08
fix: make the farcaster-related hooks work with the new API

BrickheadJohnny:


src/requirements/Farcaster/hooks/useFarcasterChannels.ts
```diff
@@ -1,23 +1,21 @@
+import { useFarcasterAPI } from "@/hooks/useFarcasterAPI"
+import { NEYNAR_BASE_URL } from "@/hooks/useFarcasterAPI/constants"
+import type { NeynarAPIClient } from "@neynar/nodejs-sdk"
```
dominik-stumpf, line 3: I have added the neynar sdk to include types. To prevent this next time, we could maybe use this. Although the update didn't seem like a small one.

dominik-stumpf (approved):

dominik-stumpf:
Thank you for taking the time! Looks good.

PR #1520, opened 2024-10-09
fix: remove old neynar api call

dominik-stumpf:
### Objective

~~Until the new neynar endpoint is finished, remove the neynar call that causes error on profiles with farcaster~~
Use `guild.xyz` without subdomain for url when requesting neynar.

### Investigation

Error was caused by having a real href swapped by the new neynar route handler (`api/neynar`), which is required by URL constructor

```tsx
const neynarRequest = fcProfile && new URL(`${NEYNAR_BASE_URL}/user/bulk?fids=${fcProfile.fid}`, api)
```

[discord reference](https://discord.com/channels/697041998728659035/1192824841703784498/1293127450141786162)

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,5 @@
 NEXT_PUBLIC_API=https://api.guild.xyz/v1
+NEXT_PUBLIC_DOMAIN=https://guild.xyz
 NEXT_PUBLIC_DISCORD_CLIENT_ID=868172385000509460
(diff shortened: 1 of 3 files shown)


TomiOhl (approved):
Talked about it irl, not that many changes & currently some pages are broken so let's merge it

PR #1539, opened 2024-11-04
impro: copiable correlation id indicator

FBalint:


src/v2/components/CopyableCorrelationId.tsx
```diff
@@ -0,0 +1,43 @@
+import { Check, Copy, X } from "@phosphor-icons/react"
```
BrickheadJohnny, line 1: Let's use the SSR icons here ;) (just change the import to `@phosphor-icons/react/dist/ssr` & specify the `weight="bold"` attribute on the icons if needed)

src/v2/components/CopyableCorrelationId.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { useCopyToClipboard } from "usehooks-ts"
+
+export const CopyableCorrelationId = ({
+  correlationId,
+}: { correlationId: string }) => {
+  const [, copyToClipboard] = useCopyToClipboard()
+  const [isCopied, setIsCopied] = useState(false)
+  const [hasFailed, setHasFailed] = useState(false)
+
+  const handleCopy = () => {
```
BrickheadJohnny, line 12: We have a custom `useCopyToClipboard` hook in `/src/v2/hooks`, let's use that here

BrickheadJohnny (commented):
Great work, thanks, I only left 2 minor comments!

BrickheadJohnny (approved):

PR #1548, opened 2024-11-18
adjust and test project setup

dominik-stumpf:
- dark theme
- github action fix
- package adjustment
- rename `cn` file


src/lib/css-utils.ts
BrickheadJohnny, line 1: I have mixed feelings about this, we probably won't really have other CSS related utils, so I'd just leave it as cn, wdyt?

src/components/csr-providers.tsx
BrickheadJohnny, line 1: I'd call this just `Providers`

BrickheadJohnny (commented):
Left 2 comments, let's discuss those later today! 😉

src/lib/css-utils.ts
BrickheadJohnny: Talked about this IRL, `css-utils` is a more descriptive name, so let's keep that!

PR #1549, opened 2024-11-19
feat: Dialog, Drawer and ResponsiveDialog components

BrickheadJohnny:


dominik-stumpf (changes requested):
I suggest using `ScrollArea` to avoid out of place native style.

image

The drawer doesn't seem to have scrollable body, but if it had would pose the problem of making it hard to dismiss using a down swipe gesture. We should think about a solution for this.

dominik-stumpf:
Found a nice example for scrollable drawer: https://vaul.emilkowal.ski/default#scrollable

BrickheadJohnny:
Thanks, nice catch, @dominik-stumpf!
Just pushed a commit where I use a `ScrollArea` component. I also removed the complex logic from the `Dialog` component for the scrollable body. For now, the body is always scrollable (if the content would overflow), if we'll need a scrollable modal later, we'll implement it.

PR #1550, opened 2024-11-19
feat: add background scale animation

dominik-stumpf:
Add drawer background animation, same as on the documentation landing: https://vaul.emilkowal.ski/


dominik-stumpf:
Okay, closing for that reason.

PR #1551, opened 2024-11-19
feat: simple sign in flow

BrickheadJohnny:
Please let me know if you'd do the server actions in a different way!


dominik-stumpf (commented):
Solid.

Here's a follow up PR for restructuring and testing the auth: https://github.com/guildxyz/guild.xyz/pull/1552

> There is a logout endpoint for v3, not sure if we should call it or not when signing out.

PR #1555, opened 2024-11-21
add explorer

dominik-stumpf:
Mostly adapted from v2 explorer, went through some simplification, replaced SWR with react-query. In this version the header can be glitched if the page is scrolled rapidly. 

- **chore: add explorer scaffolding**
- **feat: add infinite scroll**


src/app/explorer/components/InfiniteScrollGuilds.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const guilds = data?.pages.flatMap((page) => page.items) || [];
+
+  return (
+    <section className="grid gap-2">
+      <div className="grid gap-4 sm:grid-cols-2 lg:grid-cols-3">
+        {isLoading
+          ? Array.from({ length: 33 }, (_, i) => (
+              <Card key={i}>
+                <Skeleton className="size-full h-[114px]" />
```
BrickheadJohnny, line 53: Let's use the detailed guild card skeleton which I sent to you, it looks better in my opinion 😉

BrickheadJohnny (changes requested):
Love these changes, I left a couple of comments & I'd also bring back the sticky header as I said you today morning 😉

guild.d.ts
```diff
@@ -0,0 +1,33 @@
+// dumping here types util it comes down properly
+
+export {}
+
+declare global {
```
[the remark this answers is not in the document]
dominik-stumpf: I'm not certain how types will be. I would just keep them here until we generate them from the backend. This way we wont have deal with the import renamings later.

src/app/explorer/components/InfiniteScrollGuilds.tsx
```diff
@@ -0,0 +1,82 @@
...
+          },
+          [setIntersection],
+        )}
+        aria-hidden
+      />
+      {guilds.length === 0 &&
+        !isLoading &&
+        search &&
+        `No results for "${search}"`}
+      <Button
```
BrickheadJohnny, line 69: I think we can remove this now that the infinite scroll works?

src/app/explorer/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const PAGE_SIZE = 24;
+export const ACTIVE_SECTION = {
```
dominik-stumpf, line 2: Do we want these kind of consts to be in screaming snake case?

PR #1556, opened 2024-11-22
feat: create guild page

BrickheadJohnny:
Left a couple of TODO comments for later, but I think we could merge this PR now. Let me know if you notice any issues or if you have any suggestions for improvements!


src/app/create-guild/components/CreateGuildFormProvider.tsx
```diff
@@ -0,0 +1,25 @@
...
+const defaultValues = {
+  name: "",
+  imageUrl: "",
+  urlName: "test",
+  contact: "",
+} satisfies CreateGuildForm;
+
+const CreateGuildFormProvider = ({ children }: PropsWithChildren) => {
+  const methods = useForm<CreateGuildForm>({
+    mode: "all",
```
dominik-stumpf, line 17: `all` might be obtrusive, how about we try `onTouched`?

src/app/create-guild/page.tsx
```diff
@@ -0,0 +1,38 @@
...
+            Begin your guild
+          </h2>
+
+          {/* TODO: <CreateGuildImageUploader /> */}
+
+          <div className="mb-8 flex flex-col gap-4">
+            <CreateGuildForm />
+          </div>
+
+          <AuthBoundary fallback={<SignInButton size="xl" />}>
```
dominik-stumpf, line 29: We should have `primary` color scheme on this for highlight or redirect users if not signed in (like in profile).

dominik-stumpf (commented):
Nice job, did an early review to catch up!

src/app/create-guild/components/CreateGuildFormProvider.tsx
BrickheadJohnny: Oooh, that's a great idea! 💯

src/app/create-guild/page.tsx
BrickheadJohnny: Yup, I left it on the default color scheme, because I couldn't really decide if I want it to be blue or indigo ;)

src/components/ConfettiProvider.tsx
BrickheadJohnny, line 1: Mostly a copy-paste from the current app, but I removed the confetti type param, since we'll only have one confetti type for now.

src/components/ImageUploader.tsx
```diff
@@ -0,0 +1,151 @@
...
+type Props = Omit<ButtonProps, "variant" | "onClick" | "onError"> & {
+  maxSizeMB?: number;
+  onSuccess?: (imageUrl: string) => void;
+  onError?: (error: string) => void;
+  onFileInputChange?: InputHTMLAttributes<HTMLInputElement>["onChange"];
+};
+
+const mbToBytes = (mb: number) => mb * 10 ** 6;
+
+export const ImageUploader = ({
```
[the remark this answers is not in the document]
BrickheadJohnny: Yup, we should absolutely move out the logic to a custom hook later, but for now I'd just leave it here until we don't need it elsewhere 😉

PR #1557, opened 2024-11-22
add static pages

dominik-stumpf:
### Objective

Add static pages with sensible typography settings, create a component for presenting markdown prose.

### Considerations

Next time we are to extend our public prose, dynamic routes should be used for generating HTML from markdown.

- **feat: move in static pages and typography**
- **feat: add public prose and static pages**


src/app/(dashboard)/layout.tsx
BrickheadJohnny, line 1: I think we don't need this page in this PR? 👀

BrickheadJohnny (commented):
I think you included an unnecessary extra file in this PR, and the headings aren't bold on the terms of use page - if you fix these 2 issues, feel free to merge the PR 😉

src/app/(dashboard)/layout.tsx
dominik-stumpf: Talked about it before, follow up PR coming where our previous `CompoundLayout` is replaced with a dashboard layout

PR #1558, opened 2024-11-23
add guild page

dominik-stumpf:
- **feat: add guild page routes**
- **chore: center join guild button**
- **feat: add requirement ui**


src/app/(dashboard)/[guildId]/fetchers.ts
Comment (dominik-stumpf) [inline src/app/(dashboard)/[guildId]/fetchers.ts:1]: The API has changed, role group id is now indexed, so we don't have to make as much requests

dominik-stumpf (commented):
Thank you for catching up with `v3-main`, left some comments, let's discuss them together.

package.json
```diff
@@ -84,6 +85,7 @@
     "postcss": "^8",
     "storybook": "^8.4.4",
     "tailwindcss": "^3.4.1",
+    "type-fest": "^4.29.1",
```
BrickheadJohnny, line 88: Do we need this? It isn't used anywhere right now 👀

src/app/(dashboard)/explorer/page.tsx
```diff
@@ -110,16 +106,16 @@ async function YourGuildsSection() {
 }
 
 async function YourGuilds() {
-  const auth = await getTokenFromCookie();
+  const auth = await tryGetToken();
   if (!auth) return;
 
-  const { items: myGuilds } = await getAssociatedGuilds({
-    userId: auth.userId,
+  const { items: associatedGuilds } = await getAssociatedGuilds({
```
BrickheadJohnny, line 112: I'd put the `tryGetToken` call inside the `getAssociatedGuilds` function, since we only need the token for this request, wdyt?

src/app/(dashboard)/[guildUrlName]/layout.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const guild = queryClient.getQueryData(
+    guildOptions({
+      idLike: guildUrlName,
+    }).queryKey,
+  );
+  if (!guild) {
+    throw new Error("Failed to fetch guild");
+  }
+  const _user = queryClient.getQueryState(userOptions().queryKey);
```
Comment (BrickheadJohnny) [inline src/app/(dashboard)/[guildUrlName]/layout.tsx:36]: Why do we need this?

src/app/(dashboard)/[guildUrlName]/components/JoinButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+      if (reader) {
+        while (true) {
+          const res = await reader.read();
+          if (res.done) break;
+          console.log(res.value);
+          //toast(res.value?.status, { description: res.value?.message });
+        }
+      }
+
+      joinGuild({ guildId: guild.data.id });
```
Comment (BrickheadJohnny) [inline src/app/(dashboard)/[guildUrlName]/components/JoinButton.tsx:54]: The name of this function is pretty misleading, it just revalidates a tag, we should rename it in my opinion

src/app/(dashboard)/[guildUrlName]/actions.ts
```diff
@@ -0,0 +1,72 @@
...
+export const joinGuild = async ({ guildId }: { guildId: string }) => {
+  // the response type might not be suitable for this fetcher
+  revalidateTag("user");
+  //return fetchGuildApiAuthData(`guild/${guildId}/join`, {
+  //  method: "POST",
+  //});
+};
+
+export const leaveGuild = async ({ guildId }: { guildId: string }) => {
+  revalidateTag("user");
```
Comment (BrickheadJohnny) [inline src/app/(dashboard)/[guildUrlName]/actions.ts:24]: Why do we need to revalidate the user before leaving the guild?

src/actions/auth.ts
```diff
@@ -70,8 +45,13 @@ export const signOut = async (redirectTo?: string) => {
...
-  const cookieStore = await cookies();
-  const authCookie = cookieStore.get(GUILD_AUTH_COOKIE_NAME);
-  return authCookie && tokenSchema.parse(jwtDecode(authCookie.value));
+export const tryGetToken = async () => {
+  const token = (await cookies()).get(GUILD_AUTH_COOKIE_NAME)?.value;
+  if (!token) throw new Error("Failed to retrieve token");
+  return token;
+};
+
+export const tryGetParsedToken = async () => {
```
BrickheadJohnny, line 54: I'm not sure if this should throw or return undefined instead. Let's talk about it later!

src/lib/fetchGuildApi.ts
```diff
@@ -0,0 +1,171 @@
...
+  requestInit?: RequestInit,
+): Promise<FetchResult<Data, Error>> => {
+  if (pathname.startsWith("/")) {
+    throw new Error("`pathname` must not start with slash");
+  }
+  if (pathname.endsWith("/")) {
+    throw new Error("`pathname` must not end with slash");
+  }
+
+  const url = new URL(`api/${pathname}`, env.NEXT_PUBLIC_API);
```
BrickheadJohnny, line 65: Shouldn't `api/` be part of the `NEXT_PUBLIC_API` env var instead?

src/lib/fetchGuildApi.ts
```diff
@@ -0,0 +1,171 @@
...
+  const response = await fetch(url, {
+    ...requestInit,
+    headers: {
+      ...(requestInit?.body && { "Content-Type": "application/json" }),
```
BrickheadJohnny, line 69: I removed this from the previous fetcher, because we might need to send FormData to the API later (if we move the Pinata upload functionality there). Maybe we should check if the body is not FormData, and only then apply the `Content-Type: "application/json"` header?

src/lib/fetchGuildApi.ts
```diff
@@ -0,0 +1,171 @@
...
+      ...requestInit?.headers,
+    },
+  });
+
+  const contentType = response.headers.get("content-type");
+
+  if (!contentType?.includes("application/json")) {
+    throw new Error("Guild API failed respond with json");
```
BrickheadJohnny, line 77: failed **to** respond?

src/lib/fetchGuildApi.ts
```diff
@@ -0,0 +1,171 @@
...
+ * @throws If the response JSON cannot be parsed.
+ * @throws If the JWT token cannot be retrieved.
+ */
+export const fetchGuildApiAuth = async <Data = object, Error = ErrorLike>(
+  ...[pathname, requestInit = {}]: Parameters<typeof fetchGuildApi>
+): Promise<FetchResult<Data, Error>> => {
+  const token = isServer
+    ? await tryGetToken()
+    : getCookieClientSide(GUILD_AUTH_COOKIE_NAME);
+  if (!token) {
```
BrickheadJohnny, line 147: What happens when we need to send fetch requests to an endpoint both with and without auth? E.g. when we fetch the requirements/rewards for a guild? In that case, it isn't relevant if we send the token or not, we just want to display the roles which the user can see. We shouldn't throw in that case. 🤔 Or do we want to handle these requests separately? We should brainstorm about this!

BrickheadJohnny (commented):
Checked the code just to catch up & left some comments, let's talk about them today 😉

src/lib/fetchGuildApi.ts
dominik-stumpf: In this case we would have to write out `api/` in the first param, since URL truncates the second param's pathname.

src/lib/fetchGuildApi.ts
dominik-stumpf: The `Auth` fetcher has been merged with the main fetcher. Currently the token error is not propagated. This could lead to the limitation of not being able to request something as an unauthenticated user.

src/lib/fetchGuildApi.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+  if (pathname.endsWith("/")) {
+    throw new Error("`pathname` must not end with slash");
+  }
+  const url = new URL(`api/${pathname}`, env.NEXT_PUBLIC_API);
+
+  let token: string | undefined;
+  try {
+    token = await tryGetToken();
+  } catch (_) {
```
dominik-stumpf, line 69: Currently the token error is not propagated and it probably won't be later to keep the auth automatic during fetching.

src/actions/token.ts
BrickheadJohnny, line 1: I think at this point, this isn't a server action, so we can just move it to utils. What do you think?

PR #1559, opened 2024-11-25
feat: Toast component with Sonner

BrickheadJohnny:
I wasn't sure if we should define custom styles for the toast icon (fixed size, for example), let me know what's your opinion on that!


dominik-stumpf (commented):
Great job overall, I have 2 details that might need to be addressed.

- On keyboard navigation the toast doesn't indicate that it is in focus (similar how we did with modals earlier) a `ring` with `focus-visible` should do the trick.
- Nit: the close button on the top left corner overflows, aesthetically a button would do better I think on the right, where the action is.

PR #1561, opened 2024-11-25
refactor: simplify the image uploader

dominik-stumpf:
Replaced form error message on image uploader with toast as it's also a request like form submit. We could also remove the `FormErrorMessage` and use the toast inside the `ControlledImageUploader`. Let me know your thoughts on this one.

- **fix: address typo in request pathname**
- **feat: add option for toast instead of form message**


BrickheadJohnny (approved):
Thanks for the PR, I liked the toast so much that I actually just completely removed the controlled image uploader and added the toast inside the simple uploader's error handler - let's just use this one until we don't need a more complex solution!

PR #1567, opened 2024-11-26
feat: Guild, Role, RoleGroup schemas

BrickheadJohnny:
Added some basic schemas for Guilds, Roles, RoleGroups.

### DIFF
--- a/src/lib/schemas/role.ts
+++ b/src/lib/schemas/role.ts
@@ -15,7 +15,7 @@ export const CreateRoleSchema = z.object({
       logic: "AND",
       anyOfNum: 1,
     }),
-  groupId: z.number().nullish(),
+  groupId: z.string().uuid(),
 });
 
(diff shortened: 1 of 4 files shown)


src/lib/schemas/role.ts
```diff
@@ -0,0 +1,30 @@
...
+  groupId: z.number().optional().nullable(),
+});
+
+export type CreateRoleForm = z.infer<typeof CreateRoleSchema>;
+
+const RoleSchema = CreateRoleSchema.extend({
+  id: z.string().uuid(),
+  createdAt: DateLike,
+  updatedAt: DateLike,
+  memberCount: z.number().positive(),
```
dominik-stumpf, line 27: It could be nonnegative.

dominik-stumpf (approved):
I like how it follows the format of V2. We might want a comment about its intended place, which is V3.

PR #1568, opened 2024-11-26
feat: use Bun in the `code-check` GitHub Action

BrickheadJohnny:
Since we're using it for local development, I guess we can try to use it here too.


dominik-stumpf (approved):
Fast, as it should be. :rocket:

PR #1572, opened 2024-11-28
feat: requirements

BrickheadJohnny:
Added a `RequirementDisplayComponent`, which either displays a default fallback for each requirement or renders a dynamic requirement component based on the config (defined by the creator of the integration).

I also added incomplete Role & Rule schemas, we can move them out to our types package later if needed. Note, that the requirement list is not finished yet, but we don't have proper rule groups yet, so I'll make a proper design for those once our API will support them.


src/components/requirements/RequirementDisplayComponent.tsx
```diff
@@ -0,0 +1,147 @@
...
+      return (
+        <RequirementLink
+          key={node.id}
+          href={convertTemplateText(node.href, requirement)}
+        >
+          {node.value}
+        </RequirementLink>
+      );
+    default:
+      return null;
```
dominik-stumpf, line 65: Throw meaningful error here for unsupported / unimplemented requirement type. Ideally I would have the FE implemented requirement types in an array and maybe construct an object out of it with `FunctionComponent` callbacks.

src/app/(dashboard)/[guildUrlName]/[pageUrlName]/page.tsx
```diff
@@ -18,7 +20,7 @@ const GuildPage = async ({
     body: JSON.stringify({ ids: guild.pages?.map((p) => p.pageId!) ?? [] }),
   });
   const page = pages.find((p) => p.urlName === pageUrlName)!;
-  const roles = await fetchGuildApiData<Schemas["Role"][]>("role/batch", {
```
Comment (dominik-stumpf) [inline src/app/(dashboard)/[guildUrlName]/[pageUrlName]/page.tsx:21]: Why use custom schema?

src/components/requirements/RequirementDisplayComponent.tsx
```diff
@@ -0,0 +1,147 @@
...
+  RequirementImage,
+} from "./Requirement";
+
+import type { Rule } from "@/lib/schemas/rule";
+import { shortenHex } from "@/lib/shortenHex";
+import { Fragment } from "react";
+import { DataBlockWithCopy } from "./DataBlockWithCopy";
+
+const PLACEHOLDER_REGEX = /\{\{([^{}]+)\}\}/g;
+const ADDRESS_REGEX = /^0x[a-f0-9]{40}$/i;
```
dominik-stumpf, line 18: This looks like a common thing, we could move this out.

dominik-stumpf (changes requested):

src/components/requirements/RequirementDisplayComponent.tsx
BrickheadJohnny: These are just requirement UI nodes, not requirements 😉 In theory, we shouldn't run into the default case, but I added a fallback badge now.

For requirements, the worst case scenario is when we don't have a proper config in the `ui` prop, but in that case, we just fall back to the default requirement UI.

dominik-stumpf (approved):
Thank you!

PR #1574, opened 2024-11-29
feat: RewardCard component

BrickheadJohnny:


src/components/rewards/RewardCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+  image?: ReactNode | string;
+  title: string;
+  description?: string;
+  className?: string;
+}>) => (
+  <Card
+    className={cn("flex flex-col gap-4 border-2 p-4 shadow-none", className)}
+  >
+    <div className="grid grid-cols-[theme(space.9)_1fr] items-center gap-x-2 gap-y-0">
+      <div className="row-span-2 flex size-9 shrink-0 items-center justify-center overflow-hidden rounded-full bg-blackAlpha dark:bg-blackAlpha-hard">
```
dominik-stumpf, line 23: We might want to break after this on small devices so the image takes the full width of a row. Ideal place for a container query? :yolo:

dominik-stumpf (approved):
Nice. Changing the title, description and image to an individual component would be necessary for customization, but this should do until then.

src/components/rewards/RewardCard.tsx
BrickheadJohnny: Well, yep, I didn't want to optimize it for every screen size until we don't see it in action hehe. Will do it later!

PR #1578, opened 2024-12-04
chore: extract changes from `add-guild-page` branch

dominik-stumpf:
Changes extracted from during the making of explorer.

Notable changes:

- Rewritten fetcher that includes auth automatically
- Refactors - cookie handling, renamings
- Fixes to how tanstack query was setup
- Created options (utils) for data validation and fetching using tanstack query
- Some server actions were moved to client
- `NavMenu` fix: close popover on navigation


src/app/(dashboard)/explorer/constants.ts
```diff
@@ -1,5 +1,5 @@
 export const PAGE_SIZE = 24;
 export const ACTIVE_SECTION = {
-  yourGuilds: "your-guilds",
+  associatedGuilds: "associated-guilds",
```
BrickheadJohnny, line 3: tbh, I wouldn't rename this, imo "Your guilds" is much more user friendly ;)

src/app/(dashboard)/explorer/page.tsx
```diff
@@ -101,25 +99,25 @@ async function YourGuildsSection() {
...
-
-  const { items: myGuilds } = await getAssociatedGuilds({
-    userId: auth.userId,
-  });
+async function AssociatedGuilds() {
+  let associatedGuilds: Schemas["Guild"][];
+  try {
+    associatedGuilds = (await getAssociatedGuilds()).items;
+  } catch {
+    return;
```
BrickheadJohnny, line 115: We could display an error message here later

BrickheadJohnny (commented):
Left 2 minor comments, otherwise LGTM!

src/app/(dashboard)/explorer/page.tsx
BrickheadJohnny: And I'm not sure if we should fetch these guilds this way, or we should just use TanStack here too (so prefetching on server side => passing the data to the query). This way we could still render these server side request time, but we could also update the "your guilds" section when the user creates a new guild, deletes an existing one, or leaves/joins a new one)

src/app/(dashboard)/explorer/constants.ts
dominik-stumpf: Internally we started calling it like this since "your" implicates ownership, I agree that it's more user friendly.

Associated is a broader term where the logic behind it could be changed, without renaming. I'm merging this PR to move on, but we can figure out something for this.

src/app/(dashboard)/explorer/page.tsx
dominik-stumpf: > We could display a proper error message here

Error now propagates.

PR #1579, opened 2024-12-06
feat: rewards

BrickheadJohnny:


src/lib/schemas/guildReward.ts
BrickheadJohnny, line 1: This file is not 100% accurate, we'll have more rewards of course! We just included the rewards which we'd like to implement soon.

PR #1612, opened 2025-04-07
Fix husky for certain setups with nvm

TomiOhl:
The pre-push hook never worked for me. This is how we solved it on backend a year ago, thought I would commit it here


dominik-stumpf (approved):
:rocket:

